PR #3, opened 2019-02-18
A bit of context from the Truss Tech Focus document

garrettqmartin8:
Most of what @xtine did was in the [tech focus document](https://docs.google.com/document/d/1qc4wWxzG4DSNQTVNa-8x_J3DawMRpoacg8qgcXiuiv8/edit#), but I've added a few more things around version control and agile process.

### DIFF
--- a/developing/cycle/tracker.md
+++ b/developing/cycle/tracker.md
@@ -3,5 +3,5 @@
 ## Overview
 We use Pivotal Tracker to track tasks and estimate how much time each task will take.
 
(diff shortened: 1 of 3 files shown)


developing/cycle/tracker.md
```diff
@@ -0,0 +1,7 @@
+# [Development Cycle](./README.md) / Pivotal Tracker
+
+## Overview
+We use Pivotal Tracker to track tasks and estimate how much time each task will take.
+
+* [How to organize a Pivotal project](https://docs.google.com/document/d/1abmxQdYol6Zi3kR3U__JNJl3uPdKF9WPYWfu75AkZ50/edit#heading=h.yyb7wuihugn8)
```
xtine: If this is an internal Truss GDoc, I would suggest putting 🔒 next to it to designate it as such.

garrettqmartin8: Good call.

PR #5, opened 2019-02-21
Getting started with infra

garrettqmartin8:
This is based on the [work](https://www.pivotaltracker.com/story/show/163961407) @ralren did to get me up to speed on AWS and Lambda functions.


infra/getting_started/your_first_lambda_function.md
```diff
@@ -0,0 +1,185 @@
...
+    resource "aws_cloudwatch_event_rule" "every_hour" {
+      name = "every-hour"
+      description = "Fires every hour"
+      schedule_expression = "rate(1 hour)"
+    }
+    ```
+2. Specify a target, our Lambda function.
+    ```
+    resource "aws_cloudwatch_event_target" "run_test_lambda_every_hour" {
+      rule = '${aws_cloudwatch_event_rule.every_hour.name}'
```
ralren: Oh wait, this should be in double quotes.

infra/getting_started/your_first_lambda_function.md
```diff
@@ -0,0 +1,185 @@
...
+      target_id = "test_lambda"
+      arn = "${aws_lambda_function.test_lambda.arn}"
+    }
+    ```
+3. Make sure CloudWatch has permission to invoke the Lambda function.
+    ```
+    statement_id = "AllowExecutionFromCloudWatch"
```
ralren: Is this in a resource or data of some kind?

infra/getting_started/your_first_lambda_function.md
garrettqmartin8: Good catch.

infra/getting_started/your_first_lambda_function.md
garrettqmartin8: Oops! I left out the resource statement.

infra/getting_started/your_first_lambda_function.md
```diff
@@ -0,0 +1,187 @@
...
+      principal = "events.amazonaws.com"
+      source_arn = "${aws_cloudwatch_event_rule.every_hour.arn}"
+    }
+    ```
+4. Run our Terraform commands form the command line: `terraform plan` and `terraform apply`.
+
+5. Your function will now run every hour!
+
+## Cleaning up
+We've been hard at work setting up our Lambda function, but let's face it; we probably won't be relying on it as part of our production infrastructure. This is a good opportunity to learn about taking down infrastructure with Terraform. __Make sure you are inside the correct directory when running the following commands!__
```
xtine: Grammar nitpick, I would suggest — instead of ;

garrettqmartin8:
@ralren @xtine I've resolved the issues you brought up. Let me know if there's anything else I should adjust.

infra/getting_started/your_first_lambda_function.md
```diff
@@ -0,0 +1,187 @@
...
+
+    func main() {
+        lambda.Start(hello)
+    }
+
+    ```
+
+    Let's dissect this program for a second.
+
+    As you can see, we imported the AWS library we installed. We then utilize it to call `lambda.Start()`. This takes a handler function, effectively calling it, and passes Lambda requests through it.
```
ntwyman: Not actually calling the function at this point, i.e. "effectively calling it" is a little misleading. It will get called but only once a request comes in.

infra/getting_started/your_first_lambda_function.md
```diff
@@ -0,0 +1,187 @@
...
+        lambda.Start(hello)
+    }
+
+    ```
+
+    Let's dissect this program for a second.
+
+    As you can see, we imported the AWS library we installed. We then utilize it to call `lambda.Start()`. This takes a handler function, effectively calling it, and passes Lambda requests through it. 
+
+    The handler in this case is our function `hello`. This handler is relatively simple. It doesn't take any internal Lambda requests and just returns the classic "Hello, world!".
```
ntwyman: What is an "internal lambda request"?

infra/getting_started/your_first_lambda_function.md
```diff
@@ -0,0 +1,187 @@
...
+
+7. All that's left is to transform our Go program into something Lambda can actually run. We can accomplish this by creating a binary and storing it in a `.zip` file.
+
+    ```
+    GOOS=linux GOARCH=amd64 go build -o main main.go
```
ntwyman: Might be worth dissecting this a little bit, e.g. explaining  why `GOOS=linux`.

infra/getting_started/your_first_lambda_function.md
```diff
@@ -0,0 +1,187 @@
...
+
+    provider "aws" {
+        region = "${var.aws_region}"
+    }
+    ```
+
+4. Let's create that Lambda function. Add the following lines to the file.
+
+    ```
+    resource "aws_lambda_function" "test_lambda" {
```
ntwyman: Does this (and the other resources) follow the naming described in the `infra/aws/naming.md` file? If not, why not? Not saying the naming doc is right (I pulled most of it out of the air) but we should aim to be consistent

infra/getting_started/your_first_lambda_function.md
```diff
@@ -0,0 +1,187 @@
...
+        role             = "${aws_iam_role.iam_for_lambda_tf.arn}"
+        handler          = "main"
+        source_code_hash = "${base64sha256(file("main.zip"))}"
+        runtime          = "go1.x"
+    }
+    ```
+
+    `filename` points to where our `.zip` file is.
+
+    `function_name` is the name given to the function. We can invoke our Lambda function by calling this name.
```
ntwyman: this is not clear to me - the function was called `hello`, where does `test_lambda` come from?

infra/getting_started/your_first_lambda_function.md
```diff
@@ -0,0 +1,187 @@
...
+
+    `role` is the IAM role attached to this function. We will be creating this in the next step.
+
+    `handler` is the entrypoint into the function. In this particular case, it's the `main.go` file in `.zip`.
```
ntwyman: It's not the `main.go` file, it's the `main` executable go buiilt

infra/getting_started/your_first_lambda_function.md
```diff
@@ -0,0 +1,187 @@
...
+
+    `runtime` specifies what language this function will be running in.
+
+    `source_code_hash` is used to trigger updates when your source code changes.
+
+5. What's left to do finish `main.tf` is creating that IAM role.
+    ```
+    resource "aws_iam_role" "iam_for_lambda_tf" {
```
ntwyman: Again - can you check that the resource names match the naming doc and update whichever is incorrect please?

infra/getting_started/your_first_lambda_function.md
```diff
@@ -0,0 +1,187 @@
...
+6. Save the file.
+7. Back on the command line, it's time to initialize the setup for your Terraform plan.
+   ```
+   terraform init
+   ```
+8. If that's looking good, let's check out what our plan looks like.
+    ```
+    terraform plan
+    ```
+9. Provided there's nothing unsurprising in the plan, it's time to deploy your function!
```
ntwyman: What would be surprising? I have no idea what to expect from `terraform plan` so I am not really sure what would be surprising.

infra/getting_started/your_first_lambda_function.md
```diff
@@ -0,0 +1,187 @@
...
+    ```
+    terraform apply
```
ntwyman: oooh - what does it do. Unfortunately I'm on a plane, otherwise I'd try it out,,,

infra/getting_started/your_first_lambda_function.md
```diff
@@ -0,0 +1,187 @@
...
+    ```
+
+## Invoking your Lambda function
+1. To first check if you deployed successfully, let's print out a list of the existing Lambda functions.
+    ```
+    aws lambda list-functions
```
ntwyman: How does this know to look at us-west-2?

infra/getting_started/your_first_lambda_function.md
```diff
@@ -0,0 +1,187 @@
...
+   ```
+
+3. Congrats, you deployed your first Lambda function with Terraform!!!
+
+## Bonus: Set up a recurring schedule for your Lambda Function
+Let's say you want your shiny new Lambda function to run every day. Maybe every hour! To do this, you will need to use another AWS service called [CloudWatch](https://aws.amazon.com/cloudwatch/). Let's open our `main.tf` up again and add a few things:
+
+1. Add a CloudWatch rule that specifies how often the event will trigger. [This article](https://docs.aws.amazon.com/AmazonCloudWatch/latest/events/ScheduledEvents.html) has more info on how to use the `schedule_expression` value.
+    ```
+    resource "aws_cloudwatch_event_rule" "every_hour" {
```
ntwyman: That's a super greedy name that I am sure does not follow the naming guides.

infra/getting_started/your_first_lambda_function.md
garrettqmartin8: The example we use should have a clearer method signature that accepts a context argument, like the Golang Lambda handler [example](https://docs.aws.amazon.com/lambda/latest/dg/go-programming-model-handler-types.html):

```go
// ...
func hello(ctx context.Context) (string, error) {
    // ...
}
```
Then we could improve the explanation like this:

> The handler in this case is our function `hello`. This handler is relatively simple, and doesn't use the Lambda context passed to it by `Lambda.start()` or make any other external requests. It only prints, "Hello, world!".

@ntwyman I have separate question: what's the golang convention for referring to function names in prose explanations? Above, I use `hello`, but I've also seen `hello()`, which probably makes it clearer that it's a function call and not another identifier. I'm coming from Ruby where we use `ClassName#instance_method` and `ClassName.class_method`.

infra/getting_started/your_first_lambda_function.md
garrettqmartin8: I'll make sure everything is consistent!

infra/getting_started/your_first_lambda_function.md
garrettqmartin8: I think we pulled this from the Terraform docs. I'll make sure everything lines up with what is in the AWS naming doc.

infra/getting_started/your_first_lambda_function.md
garrettqmartin8: Probably nothing would be surprising, so I'll rework this.

infra/getting_started/your_first_lambda_function.md
garrettqmartin8: `aws lambda list-functions` seems to return functions for all regions, but you can use the command line flag `--region=us-west-2` to specify. I'll add that for clarity.

infra/getting_started/your_first_lambda_function.md
```diff
@@ -0,0 +1,187 @@
...
+* You have awscli configured with you credentials.
+* You know some Terraform basics like "plan" and "apply".
+
+## Creating the Go program
+1. First things first, open up that command line.
+2. Create a directory wherever you desire and name it `hello-lambda`.
+3. Set your current directory to `hello-lambda`.
+4. AWS has kindly provided a Go library to develop Lambda functions. Let's install it.
+
+    ```
```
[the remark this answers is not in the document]
garrettqmartin8: I'm seeing the markdown linter complaining now &mdash; I forgot to run `pre-commit install`.

PR #16, opened 2019-05-15
document docker config.json

ahobson:


sojeri:
so many force pushes!

developing/docker/README.md
```diff
@@ -87,3 +87,26 @@ unused networks:
...
+      "detachKeys": "ctrl-@,ctrl-["
+    }
+
+### Credentials Store
+
+You can store docker credentials in an [external credentials
+store](https://docs.docker.com/engine/reference/commandline/login/#credentials-store).
+
+On macOS, the keychain will be used by default, but if you need/want
+to store shared credentials, [pass](https://www.passwordstore.org/)
```
sojeri: is this a recommended tool over aws vault, 1password, or other tools someone might be using to store secrets?

if so, it should probably be documented at a higher level, not necessarily in the docker page here.

sojeri:
what story is this related to?

ahobson:
> what story is this related to?

It's not related to a story.  I remembered some helpful docker hints and after I shared them in slack thought I should document them more permanently.

developing/docker/README.md
ahobson: No recommendation intended, just pointing out capabilities of docker that could prove useful depending on the problem you are trying to solve.

rpdelaney: 1password has a commandline utility that can be installed with `brew install 1password-cli`. Can we suggest that instead, to keep everyone on the same tool?

ahobson: Sure!  Please update the PR to document how that works with docker.  I'm just passing along information already included in the docker documentation at the link I provided.

PR #19, opened 2019-05-19
Document direnv

rpdelaney:
I added a doc illustrating how to use `direnv`.

I'm open to suggestions on where to put this if other folks have stronger opinions. The `/developing/` topic directory table of contents is mostly sub-topics, but "Docker" has its own page. I thought of putting this under a "tools" sub-topic but that seemed a bit generic.

### DIFF
--- a/developing/direnv/README.md
+++ b/developing/direnv/README.md
@@ -1,0 +1,125 @@
+# [Tools and Practice](../README.md) / direnv
+
+## Overview
+
+[direnv](https://direnv.net/) is an MIT-licensed command-line tool that addresses the hassle of setting per-project environment variables.
+
+## Installation
+
+Via homebrew:
+
+     brew install direnv
+
+## Usage
+
+Create an `.envrc` file in a directory with some environment variables your project needs:
+
+```
+cat <<ENVRC > .envrc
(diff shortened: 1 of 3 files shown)


developing/direnv/README.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Installation
+
+Via homebrew:
+
+     brew install direnv
+
+## Usage
+
+Create an `.envrc` file in a directory with some environment variables your project needs:
```
sojeri: might be worth adding something re: putting this in the global gitignore

if we already have that explained under source control, then you can point back to there? :)

rpdelaney: Hmmm. Sometimes the `.envrc` has things that should be source controlled though. On MilMove they did something pretty neat to resolve that tension so I can put something in about it.

rpdelaney: @sojeri does 48ef255 help?

sojeri: sure, that addresses the issue enough for me :)

sojeri:
you could go for command line tools if you want to move this somewhere, but I think until we have a lot of files in here we don't need to worry about this

PR #20, opened 2019-05-19
Add some more vim plugins

rpdelaney:
Adds a few vim plugins that I think are likely to be useful to others.


developing/eid/README.md
```diff
@@ -28,9 +28,16 @@ List of editors that Truss engineers use (sorted alphabetically):
...
+[VimAwesome.com](https://vimawesome.com/) provides a list of vim plugins sorted by how many GitHub dotfiles repositories have integrated the plugin.
+
+* [vim-plug](github.com:junegunn/vim-plug) - Minimalist plugin manager
 * [vim-prettier](github.com:prettier/vim-prettier)
 * [vim-javascript](pangloss/vim-javascript.git)
+* [vim-go](github:fatih/vim-go) - golang development plugin for vim
 * [editorconfig](editorconfig/editorconfig-vim.git)
+* [Recover](github.com:chrisbra/Recover.vim) - Displays a diff before recovering a swap file
+* [vim-ale](github.com:w0rp/ale) - Asynchronous code linter
+* [vim-fugitive](github.com:tpope/vim-fugitive) - A Git wrapper so awesome, it should be illegal
```
sojeri: I don't know anything about vim, so my only thought here is: should this list be alphabetized or otherwise organized in some way? not sure how much we care at this point, since we're still at only 10 recs.

developing/eid/README.md
```diff
@@ -28,9 +28,16 @@ List of editors that Truss engineers use (sorted alphabetically):
 
 #### vi(m)
 
+[VimAwesome.com](https://vimawesome.com/) provides a list of vim plugins sorted by how many GitHub dotfiles repositories have integrated the plugin.
```
sojeri: this one should probably be a bullet item the same as the others.

rpdelaney: Hm. Since it's not a plugin, but a list of plugins, I thought it should be separate.

developing/eid/README.md
rpdelaney: I sorted them pseudo-alphabetically (ignoring leading `vim-` in the plugin names). How's that?

developing/eid/README.md
gidjin: maybe having a subheading for _resources_ and one for _good plugins_?

rpdelaney: I like that, but the heading above this one is "Plugin suggestions". What about something like [this](https://gist.github.com/rpdelaney/e03cef1b2fc7459609df8ff9dd79eea8)?

gidjin: That works for me

PR #23, opened 2019-05-24
Update instructions for publishing a terraform module

chrisgilmerproj:
I just published a new module to terraform repository. Updated the steps here.

### DIFF
--- a/infra/tf/README.md
+++ b/infra/tf/README.md
@@ -22,6 +22,12 @@ When you're ready to turn the prototype module into a published one, there's a f
    * `/github subscribe trussworks/your-repo-here`
    * `/github unsubscribe trussworks/your-repo-here statuses commits deployments public releases`
 3. Wire the repo up to the [terraform module registry](https://registry.terraform.io).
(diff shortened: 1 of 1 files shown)


infra/tf/README.md
```diff
@@ -22,6 +22,12 @@ When you're ready to turn the prototype module into a published one, there's a f
    * `/github subscribe trussworks/your-repo-here`
    * `/github unsubscribe trussworks/your-repo-here statuses commits deployments public releases`
 3. Wire the repo up to the [terraform module registry](https://registry.terraform.io).
```
cblkwell: Can we add a link to the Terraform docs on publishing a repo too (https://www.terraform.io/docs/registry/modules/publish.html)? Having the summary is nice, but I find it's useful in docs to link back to a primary source if folks want more info.

sojeri: This is a good point!

chrisgilmerproj: Done! in bf17d6a93d3bf6f8b6e64747c8b8401b8969d29b

PR #25, opened 2019-06-11
Add bash programming resources

rpdelaney:
Tracker story: [#166260192](https://www.pivotaltracker.com/story/show/166260192)

Adds resources and words of caution about bash programming.

### DIFF
--- a/developing/languages/BASH.md
+++ b/developing/languages/BASH.md
@@ -9,7 +9,7 @@ Best practices tl;dr:
   * For complicated tasks, a more modern language like Go or Python may be more reliable and easy to maintain.
 * Use [shellcheck](https://shellcheck.net)
 * Don't copy & paste code from Google or StackExchange unless you fully understand it
-* Keep your bash version up to date with your package manager (e.g. `brew install bash` and routinely running `bash upgrade`), or stick to whatever version your team has agreed to standardize on (e.g. `brew pin bash`). However, be aware that behavior of bash code can vary slightly between different versions of the interpreter.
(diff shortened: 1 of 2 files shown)


developing/languages/BASH.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Overview
+
+Best practices tl;dr:
+
+* Think hard before using bash. Something else is [often better](https://mywiki.wooledge.org/BashWeaknesses).
+  * For portability, [dash](http://gondor.apana.org.au/~herbert/dash/) provides a minimal POSIX shell feature set.
+  * For complicated tasks, a more mainstream language like Go or Python may be more reliable and easy to maintain.
+* Use [shellcheck](https://shellcheck.net)
+* Don't copy & paste code from Google or StackExchange unless you fully understand it
```
chrisgilmerproj: I might also suggest that we have something about using the latest version of bash instead of using the default found on mac osx (3.X). I think ubuntu ships 4.X and brew has 5.X. Regardless, scripts can behave differently in each version so if you DO use bash then you need to make sure you're using the right version for your project.

rpdelaney: Yeah, that's a problem, and you're right to emphasize it. I wrote a bullet about this under "Pitfalls" -- my concern is that if you are going to distribute the script anywhere, it is not easy to be assured that it will always be _executed_ with the same version as you have running locally, now and in the future.

There just isn't any good solution to this that I'm aware of. :\

chrisgilmerproj: You're right. I think though that our best practice is to ensure you're using the latest available version of bash on whatever system you're working on.

rpdelaney: OK fair, how does 82658e5 look?

chrisgilmerproj: Excellent :)

PR #26, opened 2019-06-18
Add markdown conversion of "How To Pivotal Tracker" doc

rpdelaney:
brainsik wrote this as a general overview for how to use Tracker for a client, and OK'd publication here.

### DIFF
--- a/developing/cycle/tracker.md
+++ b/developing/cycle/tracker.md
@@ -4,5 +4,77 @@
 
 We use Pivotal Tracker to track tasks and estimate how much time each task will take.
 
-* 🔒[How to organize a Pivotal project](https://docs.google.com/document/d/1abmxQdYol6Zi3kR3U__JNJl3uPdKF9WPYWfu75AkZ50/edit#heading=h.yyb7wuihugn8)
-* [Pivotal workflow](https://www.pivotaltracker.com/help/articles/workflow_overview/)
+- 🔒[How to organize a Pivotal project](https://docs.google.com/document/d/1abmxQdYol6Zi3kR3U__JNJl3uPdKF9WPYWfu75AkZ50/edit#heading=h.yyb7wuihugn8)
+- [Pivotal workflow](https://www.pivotaltracker.com/help/articles/workflow_overview/)
+
+*Remember, agile processes should be treated [more like guidelines than actual
+rules](https://www.youtube.com/watch?v=jl0hMfqNQ-g).*
+
+## Stories
+
+### State definitions
+
+- Started (waiting to be finished) -- The story is *actively* being worked on.
+
+- Finished (waiting to be delivered) -- The work is believed to be done. It may
+  or may not have passed code review yet.
+
+- Delivered (waiting on acceptance) -- For code, this means it's landed in
+  master and passed the build, or has gone all the way to production, whichever
+  makes sense for the story. For other work, use common sense.
+
+### Creating
+
+- Add enough information so anyone on the team can pick it up.
+
+- Include acceptance criteria (AC) to make it clear what "done" means and to
+  help the person accepting the work know what to validate.
+
+- Leave the story unestimated.
+
+- Add it to the top of the inbox (at the top of icebox).
+
+### Finishing / Delivering
+
+- Add comments, PRs, etc. to the story that will help someone else accept the work.
+
+### Accepting
+
+- Throughout the week, everyone should take time to accept other people's work.
+
+- Look at the acceptance criteria, validate they've been met, and click the
+  green button.
+
+## Weekly planning
+
+Weekly sprints begin with a planning session. It goes like this:
+
+1. Go through in progress stories
+
+   - Determine if the story is blocked.
+
+   - If a story will be blocked for an unknown amount of time, or if it's
+     unlikely it'll be unblocked this week, move it into the "blocked" section
+     of the icebox.
+
+2. Go through "blocked" section of the icebox
+
+   - If a story is no longer blocked, move it into backlog.
+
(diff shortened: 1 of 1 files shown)


developing/cycle/tracker.md
```diff
@@ -4,5 +4,77 @@
...
+
+- Delivered (waiting on acceptance) -- For code, this means it's landed in
+  master and passed the build, or has gone all the way to production, whichever
+  makes sense for the story. For other work, use common sense.
+
+### Creating
+
+- Add enough information so anyone on the team can pick it up.
+
+- Include acceptance criteria (AC) to make it clear what "done" means and to
```
sojeri: now we can leave a comment linking to this when folks don't include AC!  :)

developing/cycle/tracker.md
```diff
@@ -4,5 +4,77 @@
...
+     of the icebox.
+
+2. Go through "blocked" section of the icebox
+
+   - If a story is no longer blocked, move it into backlog.
+
+   - If a story has been blocked for more than a few weeks, delete it or move it
+     to the bottom of the icebox. Own up to the fact that it's not getting done.
+
+   - If a "really important" story has been blocked for more than a few weeks,
```
sojeri: should there be something here to document our preferences for self blocking vs blocked by another story? this last point here seems to imply the former is our usual case, whereas at least on Reserve we prefer the latter.

developing/cycle/tracker.md
```diff
@@ -4,5 +4,77 @@
...
+
+Weekly sprints begin with a planning session. It goes like this:
+
+1. Go through in progress stories
+
+   - Determine if the story is blocked.
+
+   - If a story will be blocked for an unknown amount of time, or if it's
+     unlikely it'll be unblocked this week, move it into the "blocked" section
+     of the icebox.
```
sojeri: here is probably a good place to add another point about creating another story to resolve the blocker (where appropriate)

developing/cycle/tracker.md
```diff
@@ -4,5 +4,77 @@
...
+     to the bottom of the icebox. Own up to the fact that it's not getting done.
+
+   - If a "really important" story has been blocked for more than a few weeks,
+     prioritize unblocking it: create a new chore to do the work that it will
+     take to unblock it and move it to the top of the backlog.
+
+3. Go through the inbox
+
+   - Read each story and make sure everyone understands what it means. If there
+     is missing context, acceptance criteria, etc. then add that in.
```
sojeri: this could use another point -- maybe a sub bullet? -- to document returning a story to its creator when folks reviewing don't have the context to add the AC etc themselves

developing/cycle/tracker.md
brainsik: What is "self blocking"?

developing/cycle/tracker.md
brainsik: What does it mean to "return a story"? Just reach out for clarification?

developing/cycle/tracker.md
rpdelaney: @brainsik a story that is blocked by something that isn't a story, i.e. writing something like "Get brainsik to explain this to us" instead of a story describing other work

developing/cycle/tracker.md
rpdelaney: This is a good thread since we have discussed this workflow orally in the reserve planning meetings but as far as I know, it's not documented anywhere yet.

My memory of how we worked this out is fuzzy. I _think_ the proposed workflow for a story missing context went _something_ like this:

1. Comment on the story tagging the requester, asking for the needed info
1. Self-block the story as waiting for a response
1. Wait some reasonable amount of time
1. If they don't respond, "reject" the story as described below

"Returning" a story (either because we did the above, or we think the work isn't needed anymore, or we don't think it belongs in the CWP bucket, etc):

1. Point the story 0 and deliver it
1. Leave a comment explaining that we are returning the story because of [reasons] and that if the requester disagrees they can reject the story and explain why we should do it anyway; otherwise it will be accepted in its current state
1. Wait some amount of time
1. If the requester disagrees and thinks we should do it, they can reject it
1. Otherwise, we accept the story and it's off our plate

@sojeri am I in the neighborhood?

developing/cycle/tracker.md
brainsik: This last point is really just about getting folks to own up to what it means for a story to get blocked. If things are blocked for a long time, you either need to accept that it's not getting done and pull it into the icebox, or step up and prioritize the work to unblock it. This is true regardless of how things are blocked (self or story). You shouldn't be processing the same story week after week. For example, you could put into the icebox and update the blocking story to have another AC that the blocked story gets moved to the inbox.

developing/cycle/tracker.md
sojeri: yeah, that sounds exactly right. now that you've written it down, it's a lot bigger than the sub bullet I was thinking. 🤔

sojeri: what about dumping this in a separate section like `when a story isn't ready yet` or `rejecting`

rpdelaney: I think it's a rly good idea to get this (1) codified & (2) added, but probably No.1 requires more people from reserve to be involved at least a little bit, so it feels like that's a separate story. Would you be okay with merging this in as-is if I wrote a story to add the thing about how to return a story?

rpdelaney: @sojeri☝️

sojeri: yes, that's fine

rpdelaney:
https://www.pivotaltracker.com/story/show/166918203

PR #28, opened 2019-07-01
AWS Organizations best practies

brainsik:
I've come up with these based on various things I've read and watched. We'll need to validate them through practice!

### DIFF
--- a/infra/aws/aws-organizations.md
+++ b/infra/aws/aws-organizations.md
@@ -1,0 +1,23 @@
+# AWS Organizations
+
+AWS Organizations provide a native way to manage multiple AWS accounts. They provide consolidated billing, APIs (e.g., via Terraform) for automating account creation, and the ability to apply account-wide IAM like policies.
+
+As security best practices prefer account boundaries over IAM policies as a way to limit resource access, AWS Organizations are becoming a standard part of any AWS deployment.
+
+## Best practices
+
+These are the best practices gleaned from a variety of resources. They are ordered by priority.
+
+* Severely limit access to the master account: this account can affect all others. Only folks who need to manage the organization should have access. IAM policies can provide granular access to Organizational Units (OUs) and Service Control Policies (SCPs).
+* Don't put any resources in the master account: it shouldn't be responsible for anything besides managing the organization.
+* Pick whether your SCPs will be whitelists or blacklists; avoid mixing: understanding what happens when a whitelist SCP inherits a blacklist SCP (or vice versa) is too complicated.
+  * Prefer blacklists as these allow folks to experiment with new services — staying out of the way of engineers.
+  * Use whitelists for larger organizations operating in high compliance environments. For example, this could let a large health care company only allow access to HIPAA approved services. Whitelists require more maintenance as new services need to get manually added to the whitelist as they are approved.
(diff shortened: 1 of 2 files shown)


infra/aws/aws-organizations.md
```diff
@@ -0,0 +1,23 @@
...
+* Don't put any resources in the master account: it shouldn't be responsible for anything besides managing the organization.
+* Pick whether your SCPs will be whitelists or blacklists; avoid mixing: understanding what happens when a whitelist SCP inherits a blacklist SCP (or vice versa) is too complicated.
+  * Prefer blacklists as these allow folks to experiment with new services — staying out of the way of engineers.
+  * Use whitelists for larger organizations operating in high compliance environments. For example, this could let a large health care company only allow access to HIPAA approved services. Whitelists require more maintenance as new services need to get manually added to the whitelist as they are approved.
+* Create a special OU with an SCP that denies everything. If an account gets popped or exhibits worrisome behavior you can move it under this OU so all services are blocked. Names folks use for this OU are: suspended, quarantined, isolated, etc. We should pick one and standardize.
+* Create a top-level OU that everything else is under: this makes it so you can create SCPs that apply to _all_ accounts.
```
cblkwell: Does this include that "quarantined" OU? I'm guessing no, since we probably don't want to worry about weird interactions between lists. I think that is the intent, but the wording here is slightly ambiguous. Maybe write a single line like "you will need two top-level OUs:" and then two subheadings detailing these two (the quarantine and everything else).

brainsik: In the example from Capital One, _everything_ was under a top-level OU. I had the same thought as you. I think since it's a Deny All, it doesn't really matter.

cblkwell: Ah. Maybe switch these around then? First talk about the top level OU everything else is under, and then say to create the quarantine zone under that? Either way, I think we should make a declaration about what we think the default should be -- it's okay to be opinionated, and letting folks know that this is what we think is best and why at least gives them something to choose if they have no strong feelings either way. Being ambiguous also makes it harder for someone to know what to expect and creates confusion.

brainsik: I'm not ready to have strong opinions about this. We haven't tried any of it! This is more like: "hey, I think this is what we should try" much more than "this is what we believe works best". That said, I'll try to remove some ambiguity, because that's not helping anyone. :-)

brainsik: I'm considering just removing this. I'm not actually sure what you get from doing this. AFAICT, you can move OUs around so it seems like you could just create this when you need it and put everything else under it.

brainsik: Removed this point and added clarification to the quarantined OU.

cblkwell: Having it there ahead of time makes sense. When that happens, you don't want to be figuring out how to do it. Better to have it ready to go -- it doesn't cost you anything after all.

brainsik: I dunno, I'm not convinced. I couldn't even come up with a scenario where I'd want this so it's starting to feel like it's just adding complexity. I only saw this referenced in one place, which was Capital One's infra, and, boy, is _that_ a large deployment.

I really think adding this in later is not that big of a deal. Create the OU, make the others children of it.

PR #29, opened 2019-07-05
Add technical design process overview and template

mikena-truss:
MilMove has been using a design doc template for the past few months to help facilitate technical design. After bringing it up in our architectural meeting, it seems like a good idea to bring it to the eng playbook as a company wide resource. I copied the template over here along with adding some flavor to the overview. I've chatted with a handful of folks outside of MilMove, but happy to discuss further if folks are interested!

FYI MilMove folks, I'll loop back and link this from our docs as the permanent resource once merged.

### DIFF
--- a/developing/technical-design/README.md
+++ b/developing/technical-design/README.md
@@ -1,0 +1,87 @@
+# [Tools and Practice](../README.md) / Technical Design
+
+## Overview
+
+This section provides some guidance
+for designing new features or architectural changes.
+While process may change based on project needs
+(such as existing client process or team size),
+these resources can provide a foundation when necessary.
+
+Some purposes of providing a process for technical design are to:
+
+* clarify problem statements and client value.
+* guide Engineers through breaking down complex problems.
+* facilitate collaboration amongst Engineers.
+* share information across teams.
+* empower Engineers of all levels to make and contribute to architectural decisions.
+
+## Design Documents
+
+Design documents help share and collaborate on technical decisions.
+They become necessary when a feature or architectural change
+is unclear based on individual pieces or work,
+has significant unknowns,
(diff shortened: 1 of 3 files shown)


developing/technical-design/README.md
```diff
@@ -0,0 +1,87 @@
...
+* Always define and readdress stakeholder value.
+* Discover your audience and keep them in mind while writing.
+  The document may be read by:
+  * Engineers directly responsible for implementation.
+  * Engineers on other teams who need to be aware of decisions and changes in the codebase.
+  * Clients, Product Managers, and/or Team Leads looking for a solution overview and resource considerations.
+  * Security experts assessing risk.
+* Foster inclusive collaboration.
+  * Emphasize main points,
+  but provide resources for less experienced team members
```
cblkwell: It might be worth explicitly calling out linking to external resources in this section.

developing/technical-design/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+Note that design documents are not a go-ahead for implementation,
+nor are they a substitute for agile workflow.
+Priorities may change based on resources, roadmaps, or unknown constraints.
+The initial design may be used as a measure for completion,
+but maintain an agile approach,
+readdress tradeoffs,
+and be adaptable to changes during the implementation phase.
+
+## Resources
```
gidjin: Any chance of linking to examples of design documents already written? I sometimes find it helpful to see a full example of such things. If none are available we might want to consider adding some into the engineering playbook.

mikena-truss: I want to, but I think right now they're all private docs. I will ask on MilMove if any _can_ be public and loop back

PR #33, opened 2019-09-04
first Infra ADR commit

eeeady:
Dynamike and I were talking earlier about where to document decisions we make about infra best practice. We liked the ADR process but needed a location to store this information so that we can share this across projects. So here it is.

We should pull ADRs across projects into here and try to de-duplicate to the best of our ability to make a concise timeline of our decisions about Infra practice.

### DIFF
--- a/infra/adr/README.md
+++ b/infra/adr/README.md
@@ -1,0 +1,16 @@
+# Architectural Decision Log
+
+Truss' infrascturcture practitioners have a breadth of knowedge and experience that should be
+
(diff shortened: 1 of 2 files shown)


eeeady:
Jeremy had a good thought which was to move the ADRs to the terraform layout repo, which solves one half of the problem. We still want to share ADR decisions responsibly across projects.

eeeady:
I think we should try having a Google Drive directory with ADRs from projects for archives.

PR #37, opened 2019-09-10
summarize PATH elements and add links

mdawn:
Closes https://www.pivotaltracker.com/story/show/168030905

- Summarized blog post
- Added 2 links
- Added tl;dr bullet point

Almost added a note about `GOPATH` illustrating the importance of `PATH`, but decided that was out of the scope of this story, and would probably belong in Go resources anyway.

### DIFF
--- a/developing/languages/BASH.md
+++ b/developing/languages/BASH.md
@@ -10,6 +10,7 @@ Best practices tl;dr:
 * Use [shellcheck](https://shellcheck.net)
 * Don't copy & paste code from Google or StackExchange unless you fully understand it
 * Keep your bash version up to date with your package manager (e.g. `brew install bash` and routinely running `brew upgrade`), or stick to whatever version your team has agreed to standardize on (e.g. `brew pin bash`). However, be aware that behavior of bash code can vary slightly between different versions of the interpreter.
+* Consolidate your bash profiles into a single `~/.bashrc` file to streamline machine instructions affecting your `PATH` environment variables.
 
 ## Debugging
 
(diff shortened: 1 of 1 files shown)


developing/languages/BASH.md
```diff
@@ -37,6 +38,17 @@ Shell programming is deceptively perilous (see [Bash Pitfalls](https://mywiki.wo
...
+## PATH
+
+* `PATH` is a shortcut to a certain location on your machine, which you can view by typing `echo $PATH` in the terminal. To oversimplify, a `PATH` is an environment variable which consists of a list of directories. Adding a `PATH` to your `bash` profile will give your machine direct instructions on where to look for a certain program.
+* One of several files can affect how your machine searches for a given `PATH`, listed in the order a Mac searches for them:
+
+  * `~/.bashrc` (read on every new shell, so looked at far more often)
+  * `~/.bash_profile` (read only on login)
+  * `~/.profile`
+
+* If you have multiple files affecting your `PATH`, consider consolidating them into one `/.bashrc` and deleting the others. Next, organize each `PATH` directory in your `~/.bashrc` from most-used to least-used to maximize readability.
```
monfresh: The reason for ordering by most-used is for efficiency, not readability.

developing/languages/BASH.md
```diff
@@ -45,3 +57,5 @@ Shell programming is deceptively perilous (see [Bash Pitfalls](https://mywiki.wo
 * [Bash FAQ](http://mywiki.wooledge.org/BashFAQ)
 * [Bash Pitfalls](https://mywiki.wooledge.org/BashPitfalls)
 * [Bash-Hackers Wiki](http://wiki.bash-hackers.org/)
+* [Path Variables](https://truss.works/blog/2016/2/26/engineer-how-to-access-and-edit-your-path-system-variable)
+* [CLI Navigation Shortcuts](https://www.gnu.org/software/bash/manual/bash.html#Introduction-and-Notation)
```
monfresh: This seems fine for now, and perhaps later we might add some caveats depending on your OS and CLI app. For example, the bash manual says you can move the cursor between words by using the meta key + f or b, but that doesn't work on a Mac with Terminal or iTerm out of the box. In Terminal and iTerm, the default is to use option + left or right arrow. If you want to use option + f or b, there is a checkbox in Terminal app to "Use option as Meta key" under Preferences -> Profiles -> Keyboard.

mdawn: Got it. Might want to make a story for that. This story was just about adding these two resources..

developing/languages/BASH.md
mdawn: As stated above. Good catch.

PR #38, opened 2019-09-18
Add docs for writing Go web services

mikena-truss:
## Overview
First pass at some guidelines for building web apps in Go.

This is a lot of my opinions after working on MilMove. Some of them are carried over, some are new or oppositional to MilMove code.

The goal here is to start some written conversation and move to a set of guidelines we can base projects on and iterate through. I'm trying to document various discussions I've had with folks that have not made it to be paper. I'm also being purposely opinionated to force conversation and converge on some commonalities. I don't think I'm right about this all (in fact, some feel experimental or *wrong*), and I'm hoping folks will meld some of their views into the review!

## Future Work
This is super incomplete. I'd like to get a baseline before adding more info. It will also be a group effort to keep this going. I've started to identify future work we can turn into tickets or keep track of:
- how to start/configure a server
- what handler instantiation patterns/packages to use (such as middleware)
- how to write API authentication/authorization code
- error models
- DB package usage
- logging libraries/patterns
- observability/metric reporting
- guides around SQL
- section on general web practices abstracted from Go
- query package patterns

## Why not a sample app?
My initial inclination is that it would be nice to have a sample app for projects to reference. The problem is, sample apps require lots of work to keep them modern. Engineers would have to continually revisit and refactor them to reflect evolved principles. 

Sample apps are also likely to be small and not represent client work. Broader guidelines and patterns can be incorporated at inception and throughout project work.

A sample app before written patterns also becomes somewhat of a "cart before the horse" situation. If someone wants to maintain a sample app or is building an in-house Go tool, they could use a set of patterns to build it and draw references from the code.

More comments inline.

Also note, it's probably time to add some TOC formatting to this repo

### DIFF
--- a/developing/languages/GO.md
+++ b/developing/languages/GO.md
@@ -57,7 +57,10 @@ when selecting new packages,
 highly consider standard libraries over third party dependencies.
 One of the strengths of Go
 is its core packages,
-such as http, json, and sql.
+such as
+[http](https://golang.org/pkg/net/http/),
+[json](https://golang.org/pkg/encoding/json/),
+and [sql](https://golang.org/pkg/database/sql/).
 These libraries also use vocabulary and patterns
 easily accessible via popular public Go resources,
 which are often translatable to modern programming approaches
@@ -71,6 +74,8 @@ Since the third party ecosystem is still new,
 many packages tend to be immature with little community support,
 follow opinionated patterns inconsistent with Go idioms,
 or lack long term support.
+Be aware that choosing a new package
+is likely to involve contributing to its upkeep.
 
 ## Time
 
--- a/developing/languages/GO.md
+++ b/developing/languages/GO.md
@@ -71,11 +71,20 @@ or domain areas (such as relational databases)
 to adjust and onboard.
 
 Since the third party ecosystem is still new,
-many packages tend to be immature with little community support,
+packages may have little community support,
 follow opinionated patterns inconsistent with Go idioms,
 or lack long term support.
-Be aware that choosing a new package
-is likely to involve contributing to its upkeep.
+When choosing a third party package,
+carefully weigh the cost adoption and support contributions.
+Document the decision
+and any shortcomings of a comparable standard library
+along with a rollback plan.
+
+### Third Party Packages
+
+Some examples of third party packages we've found to be helpful and stable are:
+
+* [sqlx](https://github.com/jmoiron/sqlx) for SQL querying and struct marshalling
 
 ## Time
 
--- a/developing/languages/GO.md
+++ b/developing/languages/GO.md
@@ -4,9 +4,155 @@
 
 Go aka Golang is a programming language designed by folks at Google.
 
-## Resources
+This guide provides some general resources for working in Go.
+[Web Application in Go](../../web/server/go.md)
+provides some specifics to web development.
+
+## Learning Resources
+
+### References
 
 * [The Go Programming Language Official Website](https://golang.org/)
+* [Effective Go](https://golang.org/doc/effective_go.html) (how to do things “the Go way”)
 * [Go Pointer Primer](https://github.com/trussworks/go-pointer-primer)
+* [GoDoc](https://godoc.org/) (where you can read the docs for nearly any Go package)
+* [Go wiki](https://github.com/golang/go/wiki/Learn)
+* _Book_: [The Go Programming Language](http://www.gopl.io/)
+* Advanced Testing with Go
+  [Video](https://www.youtube.com/watch?v=yszygk1cpEc)
+  and [Article](https://about.sourcegraph.com/go/advanced-testing-in-go) (great overview of useful techniques, useful for all Go programmers)
+
+### Tours/Lessons
+
+* [A Tour of Go](https://tour.golang.org) (in-browser interactive language tutorial)
+* [How to Write Go Code](https://golang.org/doc/code.html) (info about the Go environment, testing, etc.)
 * [Go by Example](https://gobyexample.com)
 * [Go Track on Exercism](https://exercism.io/tracks/go)
+
+## Testing
+
+### General
+
+* Use table-driven tests where appropriate.
+* Make judicious use of helper functions so that the intent of a test is not lost in a sea of error checking and boilerplate.
+  Use [`t.Helper()`](https://golang.org/pkg/testing/#T.Helper) in your test helper functions to keep stack traces clean.
+
+### Coverage
+
+* Always test exported functions.
+  Exported functions should be treated as an API layer for other packages.
+  Cover the expected behavior and error scenarios as a user of that API.
+* Try not to test unexported functions.
+  Unexported functions are implementation details of exported ones
+  and should not change the intended usage.
+  If you find that an unexported function is complex and needs testing,
+  it might mean it needs to be refactored as it's exported function elsewhere.
+
+## Time
+
+### Clock Dependency
+
+`time.Now()` can cause a lot of side effects in a codebase.
+One example is
+that you can't test the "current" time
+that happened in a function you called in the past
+
+For example, let's say we have the following:
+
+```go
+package mypackage
+
+import "time"
+
+func MyTimeFunc() time.Time {
+    return time.Now()
+}
+
+func TestMyTimeFunc(t *testing.T) {
+    if MyTimeFunc() != time.Now() {
+        // This will error!
+        // The time in the function and the test happen at different times
+        t.Errorf("time was not now")
+  }
+}
+```
+
+How do we test the contents of the return here?
+If we want to assert the time
+we need a way to know what `time.Now()` was when the function was called.
+
+Instead of directly using the `time` package,
+we can pass a clock as a dependency and call `.Now()` on that.
+Then in our tests, we can assert against that clock!
+The clock can be anything as long as it adheres to the `clock.Clock` interface
+as defined in the
+[facebookgo clock package](https://godoc.org/github.com/facebookgo/clock#Clock).
+We could, for example,
+make the clock always return the year 0,
+or the 2019 New Year,
+or maybe your birthday!
+In this clock package,
+there are two clocks.
+
+* The real clock where `clock.Now()` will call `time.Now()`.
+* A mock clock where `clock.Now()` always returns epoch time.
+  We'll show later how to change that!
+
+Let's look at the example above with the `clock` package.
+
+```go
+package mypackage
+
+import "fmt"
+import "time"
+
+import "github.com/facebookgo/clock"
+
+func MyTimeFunc(clock clock.Clock) time.Time {
+    return clock.Now()
+}
+
+// Then our caller
+func main() {
+    // clock.New() creates a clock that uses the time package
+    // it will output current time when .Now() is called
+    fmt.Print(MyTimeFunc(clock.New()))
+}
+```
+
+Then in our tests we can use a mock clock that freezes `.Now()` at epoch time:
+
+```go
+func TestMyTimeFunc(t *testing.T) {
+    testClock := clock.NewMock()
+    if MyTimeFunc(testClock) != testClock.Now() {
+        // both should equal epoch time, we won't hit this error
+        t.Errorf("time was not now")
+  }
+}
+```
+
+Cool, but what if I want to use a different date?
+Say my test relies on our `TestYear` constant.
+The [clock.Mock clock](https://godoc.org/github.com/facebookgo/clock#Mock)
+allows us to add durations to the clock and set the current time.
+Note that the `clock.Clock` interface does not allow this,
+it needs to happen before passing the mock clock through the interface parameter.
+
+#### Setting the Mock Clock
+
+Here's an example using the test above and setting the time to September 30 of TestYear:
+
+```go
+func TestMyTimeFunc(t *testing.T) {
+    testClock := clock.NewMock()
+    dateToTest := time.Date(TestYear, time.September, 30, 0, 0, 0, 0, time.UTC)
+    timeDiff := dateToTest.Sub(c.Now())
+    testClock.Add(timeDiff)
+    if MyTimeFunc(testClock) != testClock.Now() {
+        // both will now be September 30 of TestYear
+        // we'll pass the test again
+        t.Errorf("time was not now")
+  }
+}
+```
--- a/developing/languages/GO.md
+++ b/developing/languages/GO.md
@@ -48,6 +48,30 @@ provides some specifics to web development.
   If you find that an unexported function is complex and needs testing,
   it might mean it needs to be refactored as it's exported function elsewhere.
 
+## Packages
+
+### Prefer Standard Libraries
+
+In general,
+when selecting new packages,
+highly consider standard libraries over third party dependencies.
+One of the strengths of Go
+is its core packages,
+such as http, json, and sql.
+These libraries also use vocabulary and patterns
+easily accessible via popular public Go resources,
+which are often translatable to modern programming approaches
+in their respective areas.
+This creates an easier bridge
+for Engineers new to Go
+or domain areas (such as relational databases)
+to adjust and onboard.
+
+Since the third party ecosystem is still new,
+many packages tend to be immature with little community support,
+follow opinionated patterns inconsistent with Go idioms,
+or lack long term support.
+
 ## Time
 
 ### Clock Dependency
--- a/web/server/go.md
+++ b/web/server/go.md
@@ -82,11 +82,12 @@ If the integration tests become complex,
 it's a good indicator that the application logic for that handler
 is due for a refactor.
 
-Unit testing handlers often involves mocking service layers.
+Unit testing handlers often involves mocking
+[service layers](#services).
 From there,
 they will answer the question of:
-given a set of service layer responses,
-what response does the handler return.
+"given a set of service layer responses,
+what response does the handler return?"
 This verifies the response/request responsibilities of the handler,
 and can be used to test edge cases with little test performance impact.
 
@@ -233,7 +234,7 @@ Note that in many cases when writing Go code,
 we'd set parameters like these as unexported
 to avoid invalid state changes,
 however this isn't possible due to required exposure
-by DB/JSON packages that will use the models.
+by DB/JSON packages that will use the model's public fields.
 
 We've continued to separate any DB concerns
 to outside the model.
@@ -359,7 +360,7 @@ Our handler may look something like this:
 ```go
 package handlers
 
-type CalculateCutenessScore func(dogID uuid.UUID) (int, error)
+type CalculateCutenessScore = func(dogID uuid.UUID) (int, error)
 
 type getCutenessScoreHandler struct {
     calculateCutenessScore CalculateCutenessScore
@@ -428,7 +429,9 @@ to the `services` package.
 It returns a function that can calculate cuteness for a given dog ID.
 The dependencies, like the `Database` connector,
 are passed into the function generator and hidden from the return value.
-Notice how this piece is independent of the definition in the handler.
+
+Notice how this definition is independent of the definition in the handler
+(in a different package).
 The handler only needs to know what is required to return a score,
 it does *not* need to know how.
 Developing the implementation independently
@@ -442,8 +445,8 @@ it's segmented to the service function/package.
 Likewise, if we want to use a new DB connector,
 the dependency for the service will change,
 but the handler will be agnostic to that difference.
-Also, if we reuse the calculator elsewhere,
-it will not have to be updated sparsely across the handlers codebase.
+If we reuse the calculator elsewhere,
+changes will not have to be updated sparsely across the handlers codebase.
 
 #### Testing Services
 
--- a/web/server/go.md
+++ b/web/server/go.md
@@ -54,7 +54,7 @@ type getDogHandler struct {
 func (h *getDogHandler) handleGetDog(w http.ResponseWriter, r *http.Request) {
     // the only work here is getting params
     // the marshalling likely would be delegated to a middleware in production
-    p := dogGetParams(w)
+    p := dogGetParams(r)
 
     // offload querying to a service `fetchDog`
     dog, err := h.fetchDog(p.DogID)
@@ -370,7 +370,7 @@ type getCutenessScoreHandler struct {
 
 func (h *getCutenessScoreHandler) handleGetCutenessScore(w http.ResponseWriter, r *http.Request) {
     // like above we've deferred marshalling for this example
-    p := dogGetParams(w)
+    p := dogGetParams(r)
 
     // offload calculation to service function
     score, err := h.calculateCutenessScore(p.DogID)
--- a/web/server/go.md
+++ b/web/server/go.md
@@ -1,0 +1,495 @@
+# Web Applications in Go
+
+This section is meant to contain of some of the lessons
+Truss has had with building web applications in Go.
+It borrows from the format of [Effective Go](https://golang.org/doc/effective_go.html)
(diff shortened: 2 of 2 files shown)


web/server/go.md
```diff
@@ -0,0 +1,498 @@
+# Web Applications in Go
```
mikena-truss: I started writing this in the initial golang guide, but it seemed worth its own section. Feel free to suggest placement.

web/server/go.md
```diff
@@ -0,0 +1,498 @@
...
+
+This section is meant to contain of some of the lessons
```
mikena-truss: Personal opinion that Effective Go has been very helpful for me (and others have confirmed this). Feel free to suggest structure changes though if there's a better learning style.

web/server/go.md
```diff
@@ -0,0 +1,498 @@
...
+package handlers
+
+type getDogHandler struct {
+    // service implementation, we'll get back to this later
+    fetchDog FetchDog
+    //logger implementation may vary, we'll psuedocode for this example
+    logger Logger
+}
+
+func (h *getDogHandler) handleGetDog(w http.ResponseWriter, r *http.Request) {
```
mikena-truss: This can get refined as someone builds out handler resources. For now, I'm just using some basic `HandlerFunc` terminology. I know there are a lot of neat middleware and anonymous func patterns we can be using.

web/server/go.md
```diff
@@ -0,0 +1,498 @@
...
+    dog := Dog{}
+    dog.Name = name
+    dog.Owner = Owner
+    // Save is an external method to interface with DB passed as dependency
+    db.Save(&dog)
+    return dog, nil
+}
+```
+
+Notice how our data source is queried entirely outside of the model.
```
mikena-truss: This is one of the areas of "I'm being opinionated". I think large models caused some trouble in the MilMove codebase, and I'm pretty firm on using a different pattern. That being said, this was only tested in a few models and hasn't been tested throughout an application.

web/server/go.md
```diff
@@ -0,0 +1,498 @@
...
+    Name    string    `json:"name" db:"name"`
+    Owner     Owner
+}
+```
+
+Elsewhere in the code,
+we create a new dog and save it.
+
+```go
+package services
```
mikena-truss: Introducing `services` here might be confusing without the next section. If someone has better phrasing, I'm open to it.

web/server/go.md
```diff
@@ -0,0 +1,498 @@
...
+        transaction.Rollback()
+        return err
+    }
+    err := transaction.Save(owner)
+    transaction.Commit()
+    return nil
+}
+```
+
+Separating the database concerns from the model
```
mikena-truss: This especially stems for a "concurrency first" approach. Writing DB code/queries into the model makes them so inflexible. I understand the pattern from inheritance based models (such as Active Record), but it doesn't seem worthwhile with the interface/composition based Go patterns. I especial see this with locking updates are introduced into a codebase. The second application logic gets spliced in the middle of that isolation, writing it into the models gets heavily overloaded. It's possible this is an anti-pattern and would love some input.

web/server/go.md
```diff
@@ -0,0 +1,498 @@
...
+it becomes much harder to maintain models as purely structural.
+It's hard to make an exact recommendation for Go code structure
+as data requirements may vary considerably.
+Backends could vary from REST APIs,
+highly normalized SQL tables,
+serialized document stores,
+sparsely column oriented data,
+and so on.
+
+Some general considerations when organizing query code are:
```
mikena-truss: Would be cool to build out some links to real world examples here in the future.

web/server/go.md
```diff
@@ -0,0 +1,498 @@
...
+
+    w.Write(cutenessPayload(score))
+}
+```
+
+The handler here is not too much different than our `GetDog` example above.
+We've changed from `GetDog` to `GetCutenessScore`
+and likewise updated the service definition.
+We've also updated the error catching to be a little more robust.
+In this case,
```
mikena-truss: I'm hoping this doesn't cause confusion. I want to emphasize that the interface between handlers and services is a great way to dissect errors, rather than just pass whatever you get with a 500 attached.

web/server/go.md
```diff
@@ -0,0 +1,498 @@
...
+This is important to understanding services,
+so let's dive in a little deeper.
+
+First let's look at a possible service definition for the cuteness score.
+
+```go
+package services
+
+// NewCalculateCutenessScore return an implementation for calculating dog cuteness
+func NewCalculateCutenessScore(db Database) func(dogID uuid.UUID) (int, error) {
```
mikena-truss: This is new! Go has typed functions and I think we should make more use of them. Especially since we've defined services as single method use. This seems like a much cleaner view at dependencies vs business logic to me. It's also easier to test/pass with less boilerplate. I created a proof of concept with the MilMove codebase at one point and liked it. Also ran it by some other folks and the JS-like closure aspect was appealing to folks from other language backgrounds.

web/server/go.md
```diff
@@ -0,0 +1,498 @@
...
+}
+```
+
+So, we've added a new implementation for cuteness score
+to the `services` package.
+It returns a function that can calculate cuteness for a given dog ID.
+The dependencies, like the `Database` connector,
+are passed into the function generator and hidden from the return value.
+
+Notice how this definition is independent of the definition in the handler
```
mikena-truss: This is the meat of the definition. Please suggest clarifications if necessary.

web/server/go.md
```diff
@@ -0,0 +1,498 @@
...
+Avoid over-testing dependencies,
+and attempt to mock them when possible
+(such as DB calls),
+especially expensive or unknown resources such as third party APIs.
+These are likely to create long
+and flaky tests.
+
+#### Service Resources
+
+There  are three patterns worth understanding are:
```
mikena-truss: It might be helpful to lead with these resources. LMK.

web/server/go.md
```diff
@@ -0,0 +1,498 @@
...
+   and implementations will implicitly satisfy those contracts.
+3. [What is a Closure](https://www.calhoun.io/what-is-a-closure/).
+   Taking the "one function interface" pattern a step further,
+   we can utilize Go's first class functions and closures
+   to simply pass the function rather than struct,
+   removing a layer of indirection.
+
+### General Patterns
+
+This section is for patterns that don't fit into the specific layers defined above
```
mikena-truss: Filling this out is a TODO. Was going to fill this in, but the doc ended up being larger than I thought. Happy to delete for now.

developing/languages/GO.md
```diff
@@ -4,9 +4,184 @@
...
+* Always test exported functions.
+  Exported functions should be treated as an API layer for other packages.
+  Cover the expected behavior and error scenarios as a user of that API.
+* Try not to test unexported functions.
+  Unexported functions are implementation details of exported ones
+  and should not change the intended usage.
+  If you find that an unexported function is complex and needs testing,
+  it might mean it needs to be refactored as it's exported function elsewhere.
+
+## Packages
```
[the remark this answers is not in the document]
tinyels: This might be a good starting point: https://github.com/golang-standards/project-layout
On MilMove we have been discussing whether our physical architecture (e.g. package directory structure) should match features more closely. That is, the go equivalent of redux ducks. This would allow teams clearer ownership of features and modules in the code and make it easier to see when they are touching code that effects other teams.

developing/languages/GO.md
```diff
@@ -4,9 +4,184 @@
...
+### Coverage
+
+* Always test exported functions.
+  Exported functions should be treated as an API layer for other packages.
+  Cover the expected behavior and error scenarios as a user of that API.
+* Try not to test unexported functions.
+  Unexported functions are implementation details of exported ones
+  and should not change the intended usage.
+  If you find that an unexported function is complex and needs testing,
+  it might mean it needs to be refactored as it's exported function elsewhere.
```
reggieriser: If we want to enforce black box testing, should we have a rule that tests should be in a separate `*_test` package and not in the same package as the code being tested?  ([reference](https://stackoverflow.com/questions/19998250/proper-package-naming-for-testing-with-the-go-language))

developing/languages/GO.md
```diff
@@ -4,9 +4,184 @@
...
+for Engineers new to Go
+or domain areas (such as relational databases)
+to adjust and onboard.
+
+Since the third party ecosystem is still new,
+many packages tend to be immature with little community support,
+follow opinionated patterns inconsistent with Go idioms,
+or lack long term support.
+Be aware that choosing a new package
+is likely to involve contributing to its upkeep.
```
reggieriser: While I understand the concern here, this feels a little too opinionated to me toward the side of "only use the standard library".  I'd hate for us to get into a "not invented here" syndrome with this.  I think the big takeaway is that one should exercise caution in choosing third-party dependencies (with any language, but perhaps more so in Go), and weigh the risks you mentioned against the benefits.  For what it's worth, there was some [discussion about this topic](https://changelog.com/gotime/80#transcript-112) in a recent Go Time podcast.

developing/languages/GO.md
```diff
@@ -4,9 +4,184 @@
...
+How do we test the contents of the return here?
+If we want to assert the time
+we need a way to know what `time.Now()` was when the function was called.
+
+Instead of directly using the `time` package,
+we can pass a clock as a dependency and call `.Now()` on that.
+Then in our tests, we can assert against that clock!
+The clock can be anything as long as it adheres to the `clock.Clock` interface
+as defined in the
+[facebookgo clock package](https://godoc.org/github.com/facebookgo/clock#Clock).
```
jim: Why not just pass in a `time.Time` value for now instead of a clock instance?

web/server/go.md
jim: I am not convinced that separating the model layer from the database provides enough benefits to justify the overhead of an additional layer. I think you're going to need integration tests that test the models and database interacting as long as you have any kind of constraints in the database, which means that separating the database and the models needs to provide enough benefits to justify now needing to write an additional set of tests for the models in isolation in addition to the overhead of managing another abstraction.

IMHO applications don't change databases often enough for that possibility to drive application architecture. I think we need better justification for this recommendation.

developing/languages/GO.md
jim: I don't think that someone is likely to use a third-party JSON library at this point, but they might want to use a higher-level database abstraction than the built-in `database/sql`. How to we describe this recommendation in such a way that it can address that situation?

rswerve: Add language to the effect of: if you do use a third-party library, be prepared to explain why the standard library couldn't meet our needs, and also why you think the risks of a third-party library are justified or mitigated in this case.

developing/languages/GO.md
```diff
@@ -4,9 +4,184 @@
...
+* [A Tour of Go](https://tour.golang.org) (in-browser interactive language tutorial)
+* [How to Write Go Code](https://golang.org/doc/code.html) (info about the Go environment, testing, etc.)
 * [Go by Example](https://gobyexample.com)
 * [Go Track on Exercism](https://exercism.io/tracks/go)
+
+## Testing
+
+### General
+
+* Use table-driven tests where appropriate.
```
nelztruss: I generally like calling out (in a comment) the different sections of AAA (arrange/act/assert), especially in table-driven tests. https://medium.com/@pjbgf/title-testing-code-ocd-and-the-aaa-pattern-df453975ab80

web/server/go.md
```diff
@@ -0,0 +1,498 @@
...
+    // service implementation, we'll get back to this later
+    fetchDog FetchDog
+    //logger implementation may vary, we'll psuedocode for this example
+    logger Logger
+}
+
+func (h *getDogHandler) handleGetDog(w http.ResponseWriter, r *http.Request) {
+    // the only work here is getting params
+    // the marshalling likely would be delegated to a middleware in production
+    p := dogGetParams(w)
```
reggieriser: Nit: Parameters would be coming from the request `r` rather that the writer `w`.  One more case of this below.

developing/languages/GO.md
```diff
@@ -4,9 +4,184 @@
...
+* Make judicious use of helper functions so that the intent of a test is not lost in a sea of error checking and boilerplate.
+  Use [`t.Helper()`](https://golang.org/pkg/testing/#T.Helper) in your test helper functions to keep stack traces clean.
```
nelztruss: I've also heard of (but haven't explicitly played with it myself) people saying that every test should invoke `t.Parallel()` by default, and only NOT do that when you need to for explicitly testing timing-related code.

developing/languages/GO.md
```diff
@@ -4,9 +4,184 @@
...
 * [The Go Programming Language Official Website](https://golang.org/)
+* [Effective Go](https://golang.org/doc/effective_go.html) (how to do things “the Go way”)
 * [Go Pointer Primer](https://github.com/trussworks/go-pointer-primer)
+* [GoDoc](https://godoc.org/) (where you can read the docs for nearly any Go package)
+* [Go wiki](https://github.com/golang/go/wiki/Learn)
+* _Book_: [The Go Programming Language](http://www.gopl.io/)
+* Advanced Testing with Go
+  [Video](https://www.youtube.com/watch?v=yszygk1cpEc)
+  and [Article](https://about.sourcegraph.com/go/advanced-testing-in-go) (great overview of useful techniques, useful for all Go programmers)
+
```
nelztruss: Suggested add: Go Proverbs - https://go-proverbs.github.io/

nelztruss: Also suggestion to align happy-path to the left... https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88 (Maybe this isn't the right section for this?!?)

Ooooh! Also! No magic package globals! https://peter.bourgon.org/blog/2017/06/09/theory-of-modern-go.html

web/server/go.md
```diff
@@ -0,0 +1,498 @@
...
+
+The responsibility of a handler is to accept an HTTP request,
+delegate the work to complete the request,
+and respond based on error or success scenarios.
+Business logic, database code, etc. should not be in a handler.
+
+It should look rather declarative
+with error and success cases drawing two vertical lines via indentation.
+Here's a trimmed down example of a basic GET handler:
+
```
nelztruss: I _really_ don't know if this is the right section for this... Because it might tread into more general polymorphism behavior, but I've definitely seen it incorrectly applied to `Handler/HandlerFunc` stuff...

ALL `HandlerFunc` are `Handler`s, but all `Handler`s are not `HandlerFunc`. Return signatures should rarely be `HandlerFunc` (even if that is the underlying implementation)... They should probably be `Handler`s.

(I may be doing the wild-person arm-waving thing here... Let me know if I can try to make gooder words about this...)

web/server/go.md
nelztruss: I am _very_ in favor of keeping model and data access layers separate. (I deeply distrust the Active Record pattern.) This also matches the recommendations of the onion layer model of looking at code: https://www.codeguru.com/csharp/csharp/cs_misc/designtechniques/understanding-onion-architecture.html

web/server/go.md
```diff
@@ -0,0 +1,498 @@
...
+First, let's try to add a dog.
+
+```go
+package models
+
+type Dog struct {
+    ID      uuid.UUID `json:"id" db:"id"`
+    Name    string    `json:"name" db:"name"`
+    Owner     Owner
```
nelztruss: I'm open to being shown a better way, but in my experience, fully including another model object (`Owner`) in this model object (`Dog`) makes more advanced usages harder. I've learned to prefer if there were just a simple `OwnerID`... This makes it possible for the `Owner` to actually exist in a different data store, and for `Owner`s and `Dog`s to be cached independently.

Super open to push back on this... I could well be over-complicating and over-engineering and over-future-proofing.

web/server/go.md
```diff
@@ -0,0 +1,498 @@
...
+but we'll attempt to define some common patterns and inform future approaches.
+
+### Handlers
+
+The term "handler" is used to describe the layer that accepts HTTP requests
+and returns a response.
+While it may vary in implementation between different libraries,
+the term comes from the Go standard library [HTTP Package](https://golang.org/pkg/net/http/#Handler).
+
+The responsibility of a handler is to accept an HTTP request,
```
nelztruss: Yes to all of this.

Also, I tend to think of the `Handler`s being in the "Transport Layer". It's main responsibility is to translate from transport-native "language" to internally-used "language" used in the service layer: JSON -> biz/model objects; cookies/headers -> principal/user object; etc...

web/server/go.md
```diff
@@ -0,0 +1,498 @@
...
+to be built out.
+This layer will often be referred to as "business logic"
+or "application logic",
+but it's more or less
+what makes your application/API useful to a client.
+It's also where the various package dependencies
+of the application
+are likely to meet,
+answering the question of
+"how do I accomplish the client need with the given dependencies?"
```
nelztruss: Yes!

Also, I generally lump in the following responsibilities in the services layer: biz logic security and auth; coordination of multiple resources; etc.

Another note on security - this is the layer to be answering the question "does this principal have auth to access change this data", as opposed to doing it in the data-access layer. In my experience the data-access layer should be, and I believe by necessity, has full access to all (or at least most) of the data.

Again, I'm open to pushback on this tho!

web/server/go.md
nelztruss: I'm looking forward to the enhanced error values and wrapping in 1.13 too!

nelztruss: One thing I like to call out for things like `NewCalculateCutenessScore` is when it will be invoked... I believe this only gets called once (or maybe infrequently?) at server-startup-time... Whereas the returned `func(...)` gets called multiple times (and possibly concurrently) within the context of single web requests.

web/server/go.md
nelztruss: I have some Java-influenced(?) modeling patterns that I usually apply here, but I'm super open to challenging them.

I like to just define an interface for the service in question. Then to have an implementation-specific sub-package that actually implements it (or two...). But this gives consuming code a noun to hold, that has verbs on it. Under the covers Go lets the implementations be closures or whatever (or not) as is useful for your implementation. Also, in the service interface definition layer, I love automatically creating the parallels to `HandlerFunc`, which I believe facilitates unit testing and the closure implementations.

I ack that this sometimes feels ceremonious (a la Java) and wordy, but the flexibility have been really useful for me in the past.

web/server/go.md
nelztruss: Also, transport-layer stuff shouldn't make it down to this layer! No `http` package things should go in/out of this layer.

nelztruss: Oh, also! I would suggest all/most methods at these layers take a `context.Context` object. (Maybe this is even something that should be more generally mentioned?)

web/server/go.md
nelztruss: I'd suggest something along the lines of this (via https://peter.bourgon.org/go-best-practices-2016/): 
> Loggers are dependencies, just like references to other components, database handles, commandline flags, etc.

I've been a huge fan of the go-kit log package and the patterns it espouses... But I think those patterns may be useful/applied to other logging packages that are more performant. (Tho, that may be premature optimization.)

web/server/go.md
nelztruss: Ahhh, maybe the `context.Context` should be in the "General Patterns" section...?

developing/languages/GO.md
mikena-truss: I think it's _ok_ in some scenarios, but general pattern should be to think about why that's happening. I think long, sequential algorithms often amount to private method testing, which is reasonable. But generally, if code is getting super deep, passing in dependencies that are tested separate seems cleaner to me.

Integration tests should fall into a separate package though. This is not a distinction we've made in the past. Something that can probably be addressed later in this doc.

developing/languages/GO.md
mikena-truss: I'm not sure I understand the question. `time.Time` is an instance of time, `clock` reads time, which is most commonly used to call `.Now`.

web/server/go.md
mikena-truss: I think there is some heavy room for update here in the future. I don't have a great example of a successful handler layer. On MilMove, we use go-swagger, and I'd much prefer closer to standard library language around it.

I've delved into this a little with handler wrapper patterns: https://medium.com/@matryer/the-http-handler-wrapper-technique-in-golang-updated-bc7fbcffa702, but that's about all I've gotten to.

If you want to update that, I'm sure folks would be happy :)

web/server/go.md
mikena-truss: How would someone do a join query in this case? Would the relations live in different structs or arrays? How would it get translated to nested JSON?

web/server/go.md
mikena-truss: @jim what's the benefit of keeping these together? The models have to be passed via struct, since their fields are required for marshalling. They're also passed across multiple layers by necessity of data passing. This means all their methods are going to be passed as well and pulled in via package imports/exposed publicly. If the DB connector/interface is included, it is now essentially permanent/hard to move in the codebase due to its propagation. These seem like easy concerns to separate, since those methods aren't required to pass data.

I don't think it's common for someone to point blank change a DB per say (although, who knows if we'll work on some non-postgres projects with different data needs), but it is common to change the connector or access patterns. For example, applications moving from a single monolith talking directly to a DB to services with APIs require this sort of separation for a refactor without breaking their client contracts.

web/server/go.md
mikena-truss: ➕ to keeping `http` packages out of here. 

Authorization always seemed to go here in my head, since it is independent of the data access method. I think the next iteration of this could talk about authorization patterns. I don't have the best pattern right now, but I know MilMove is doing some work with roles and it seems like it could relate.

web/server/go.md
mikena-truss: Going to also add a note about being careful about pointers in these situations. Both for concurrency problems and memory usage.

web/server/go.md
mikena-truss: MilMove uses a similar layout here: https://github.com/transcom/mymove/tree/master/pkg/services. I don't love the templating it seems to create. I find it's both more go-like and easier to read when the functions are defined directly where they're used. It seems to avoid one or two layers of indirection. I don't know if defining them separate or using a noun has enough benefit to justify those layers. Unless it's meant to be a catalog of sorts.

I think the package structure, like whether to use one or many, is unclear as to best patterns for me. I'd like to experiment with something different.

developing/languages/GO.md
mikena-truss: Going to change the language around here to soften it a bit.

developing/languages/GO.md
nelztruss: Ooh yah. "teardown" is also a good thing to call out!

web/server/go.md
nelztruss: Not sure I'm following on the JOIN query bit... (Probably because I've been mostly in NoSQL for the last couple of years...) If you've got the `Dog.ID` and the `OwnerID`, you have everything you need, right?
As for "where" the relations live, it depends on the cardinality relations I believe. I was assuming one `Owner` has many `Dog`s, meaning the list of dogs for an owner is just an array of `ID`s... And a `Dog` has 0-1 `Owner`, which lives at `Dog.OwnerID`.
Translating to JSON would live in the "transport" layer... But I'd move away from needing to represent it that way... Opting more for a "each noun is individually retrievable (and cacheable), let the front-end and/or GraphQL sort 'em out" kind of approach.

This is probably too complex and confusing of a hand-waving thing I'm doing for this venue. Hrm... 🤔

nelztruss: Maybe I should take the "if you can't explain it well enough, you don't really understand it yourself" counsel... 😀

PR #1, opened 2019-02-01
First attempt at AWS naming guidance

ntwyman:
Was hoping to be able to reverse engineer AWS naming guidance from our TF scripts but it still seems to be actively evolving.

Not sure if this is a) premature or b) I just don't know where to look.

PR #46, opened 2019-10-02
Define our git workflow best practices

monfresh:
**Why**: To establish conventions, maintain consistency, and to have a
single place where folks can get answers.

This PR covers basic repo maintenance and hygiene, GitHub-specific
settings, git workflow, branch naming, and commit message format.

### DIFF
--- a/developing/vcs/README.md
+++ b/developing/vcs/README.md
@@ -2,7 +2,113 @@
 
 ## Overview
 
-Source control has less to do with the vcs system you use and more to do with the stages code goes through on the journey from inside a developers head, tested and merged into the main line, deployed to production and thence in time to being identified as that accursed legacy system that is causing all the problems.
+Source control has less to do with the VCS system you use, and more to do with the stages code goes through on the journey from inside a developer's head, tested and merged into the main line, deployed to production, and thence in time to being identified as that accursed legacy system that is causing all the problems.
+
+## Repo maintenance and hygiene
+
+* Avoid including files in source control that are specific to your development
+machine or process. For example, your editor or OS might generate files that are
+not pertinent to your project. Since those types of files will need to be
+[ignored](https://help.github.com/en/github/using-git/ignoring-files) from all
+the repos you work on, it's more convenient to set them once in a global
+`.gitignore` file (typically named `~/.gitignore_global`), as opposed to in each
+repo's `.gitignore`. Once you create your file and add rules to it, you can tell
+Git to use it:
+
+  ```
+  git config --global core.excludesfile ~/.gitignore_global
+  ```
+
+* Delete local and remote feature branches after merging. This allows us to
+reuse branch names, and makes it easier for you to differentiate between
+your branches that are done and those that are still being worked on. Deleting
+merged branches on GitHub should be automated with [Probot]. Locally, you can
+add an alias to your `~/.gitconfig` file. Example:
+
+  ```
+  [alias]
+    fresh = "!git co master && git pull && git branch --merged | grep -v '\\*' | xargs -n 1 git branch -d"
+  ```
+
+  Then instead of doing the typical `git checkout master && git pull`, you would
+  only run `git fresh`.
+
+[Probot]: https://probot.github.io/apps/delete-merged-branch/
+
+### GitHub repo settings
+
+* [Protect the master branch](https://help.github.com/en/articles/configuring-protected-branches)
+by turning on the following settings at a minimum:
+
+  * Require pull request reviews before merging
+  * Require status checks to pass before merging
+  * Include administrators
+
+* Under `Settings->Merge button`:
+  * Uncheck all options except for `Allow merge commits` because it's the only
+  one that allows deleting local branches with `-d`. We should never have to use
+  `-D` to delete a branch.
+
+## PR size
+
+Pull requests should be small and focused enough to be reviewable in under 15,
+or ideally under 5, minutes. Small PRs have a better chance of being quickly
+accepted. The longer the PR, the more an engineer has to keep in their head, and
+the more contiguous time they need to schedule for the review. This might lead
+to a less thorough and delayed review, which could block you from your other
+work. Long-lived PRs also often require more frequent updates with the main
+branch, which could result in mistakes when conflicts are present.
+
+## Git workflow
+
+### Fork vs clone
+
+If it's a repo you have write and push access to, clone it. Otherwise, fork it.
+
+### Cloning with SSH vs HTTPS
+
+Use [HTTPS] for read-only repos (where you’re only ever going to clone and pull),
+and use [SSH] for everything else. SSH supports certificates on security keys,
+and HTTPS does not. Therefore, we should not use HTTPS for anything that requires
+authentication. However, if you’re just pulling a public repo, HTTPS is more
+convenient since you don’t need to authenticate anything (unlike SSH for
+read-only repos).
+
+[SSH]: https://help.github.com/en/articles/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent
+[HTTPS]: https://help.github.com/en/github/creating-cloning-and-archiving-repositories/cloning-a-repository
+
+### Commit messages
+
+For each meaningful commit, write a
+[good commit message](https://chris.beams.io/posts/git-commit/) following these
+seven rules:
+
+1. Separate subject from body with a blank line
+2. Limit the subject line to 50 characters
+3. Capitalize the subject line
+4. Do not end the subject line with a period
+5. Use the [imperative mood] in the subject line
+6. Wrap the body at 72 characters
+7. Use the body to explain what and why vs. how
+
+Example:
+
+```
+Replace chromedriver-helper with webdrivers
+
+`chromedriver-helper` says in bold on their README that they are no
+longer maintaining it and that people should use `webdrivers` instead.
+
+Additionally, `webdrivers` has a very useful feature that will
+automatically use the right version of `chromedriver` based on the
+version of Chrome that is installed on the machine. This was not
+possible with `chromedriver-helper`, which made it harder to deal with
+versioning locally and in Circle CI.
+
+Resolves #123
+```
+
+[imperative mood]: https://en.wikipedia.org/wiki/Imperative_mood
 
 ## Tools
 


developing/vcs/README.md
```diff
@@ -2,7 +2,94 @@
...
+* Avoid including files in source control that are specific to your development machine or process.
+* Ignore files that are specific to your machine or process in your `~/.gitignore_global` as opposed to your repo's `.gitignore`.
+* Delete local and remote feature branches after merging.
+* Perform work in a feature branch.
+* Rebase frequently to incorporate upstream changes.
+* Use a pull request for code reviews.
+
+### GitHub repo settings
+
+* [Protect the master branch](https://help.github.com/en/articles/configuring-protected-branches) by turning on the follow settings at a minimum:
```
rpdelaney: follow settings -> following settings

developing/vcs/README.md
```diff
@@ -2,7 +2,94 @@
...
+* More information about commit (under 72 characters).
+* More information about commit (under 72 characters).
+
+Resolves #123 if GitHub issue or link to Pivotal story
+```
+
+To help enforce this commit message format, it's helpful to add our pull request
+template in your repo.
+
+If you've created more than one commit, use `[git rebase interactively](https://help.github.com/en/articles/about-git-rebase)` to squash them into cohesive commits with good messages:
```
rpdelaney: I don't think there has been a poll recently but this seems like a lightning rod to me.

monfresh: Interesting. I did not realize this was a controversial subject. I plan on expanding this to specify that this applies to noisy commits that don't add much value, such as fixing typos, lint errors, or commits made while testing a third party integration. For example, when trying to configure Circle CI, one might try different things before getting it right. I don't think it makes sense to keep all the commits that didn't work. 

Do folks have strong opinions in favor of keeping noisy commits as well?

I think that squashing into meaningful and concise commits clarifies the intent and purpose of individual commits when looking through history.

nelztruss: IME, I've usually opted for keeping the noisy stuff. Partially that was laziness, partially has to do with wanting to remind myself how the sausage was made, partially has to do with the cleanup workflow on my machine... (Not having to do a _force_ delete of the branch that I pushed up, e.g. `git branch -d examplebranch` vs `git branch -D examplebranch`...)

Yet I am not feeling super married to it, so I'm open to trying it the squash way too.

rpdelaney: > I plan on expanding this to specify that this applies to noisy commits that don't add much value, such as fixing typos, lint errors, or commits made while testing a third party integration. For example, when trying to configure Circle CI, one might try different things before getting it right. I don't think it makes sense to keep all the commits that didn't work.

I don't see any issue with rebasing commits that haven't been pushed yet. If they have been pushed, then the workflow may invove routine use of `push --force`, which is sketchy imo.

Regardless, I should have been more explicit. The lightning rod, in my view, is about rebase vs merge.

monfresh: This is great. It has given me more ideas on what to write about. There is a way to squash commits and still be able to use `-d`. As for force pushing, that would only happen once, when the PR is ready to be merged.

nelztruss:
Suggested addition: I have found this diagram to be supremely helpful whenever I've gotten myself into a bind: http://justinhileman.info/article/git-pretty/

developing/vcs/README.md
rpdelaney: > As for force pushing, that would only happen once, when the PR is ready to be merged.

I understand. I think we should have a consensus that this is a best practice before we publish it as best practice. It has been awhile since there was broad discussion about it but I believe I'm not alone in having reservations about it.

developing/vcs/README.md
```diff
@@ -4,25 +4,50 @@
...
-* Use a pull request for code reviews.
+* Ignore files that are specific to your machine or process in your
+`~/.gitignore_global` as opposed to your repo's `.gitignore`.
+* Delete local and remote feature branches after merging. This can be automated
+on the GitHub side with [Probot](https://probot.github.io/apps/delete-merged-branch/).
+Locally, you can add an alias to your `~/.gitconfig` file. Example:
+
+  ```
+  [alias]
+    fresh = "!git co master && git pull && git branch --merged | grep -v '\\*' | xargs -n 1 git branch -d"
```
rpdelaney: There is `git fetch --prune`

developing/vcs/README.md
```diff
@@ -2,7 +2,225 @@
...
+```
+git commit --amend
+```
+
+When the feature/fix is complete and tests pass, rebase to incorporate upstream
+changes.
+
+```
+git fetch origin
+git rebase origin/master
```
gidjin: A while back we had a big discussion on this flow vs merging and it never resolved. There were good points on both sides and the meeting we took a poll and it was split 50/50. I think for this document it probably makes sense to encourage merging to update your branch, and have rebase be in a separate more advanced doc, since that can lead you down some tricky paths, especially for people new to git.

developing/vcs/README.md
```diff
@@ -2,7 +2,225 @@
...
+commits you made during code review into meaningful ones:
+
+```
+git rebase -i origin/master
+```
+
+Force push the branch back up to GitHub:
+
+```
+git push --force-with-lease origin <branch-name>
```
gidjin: This is the safer option, but you can still trick --force-with-lease into overwriting stuff especially if you are using `git fetch` vs `git pull` to update your branch. Here's a helpful post that we can link here for more details https://blog.developer.atlassian.com/force-with-lease/ though I still think this is probably better as part of an advanced usage section.

developing/vcs/README.md
macrael: Yep! I much prefer `git merge` and am happy to talk about why!

monfresh: How do y'all squash commits when you use `merge` to pull in changes from `master`?

rpdelaney: Not speaking for anyone else but I don't squash pushed commits, when merging or otherwise. In general I don't think rewriting history is necessary or helpful, and the risks are significant. 

There may be distinct questions here though:

1. Is it correct to rebase a branch after code review, prior to merging (and therefore force push after rebasing)?
2. Regardless of #1, is it correct to "squash and merge" when merging to master?

I tend to think the answer to both is 'No' but my conviction about the second one is much weaker than the first. It sounds like you're asking the second question.

monfresh: Yes, I believe squashing into meaningful commits, that pass tests and other checks results in a history that's easier to read, understand, and debug. Check out my latest commit, where I add the pros and cons for both, and a list of popular repos that enforce squash and rebase: https://github.com/trussworks/Engineering-Playbook/pull/46/commits/4da0ce11c68ef2334e0a1474f59d5c0bb6b10464

developing/vcs/README.md
```diff
@@ -2,7 +2,321 @@
...
+
+## Git workflow
+
+### Fork vs clone
+
+If it's a repo you have write and push access to, clone it. Otherwise, fork it.
+
+### Cloning with SSH vs HTTPS
+
+Do we have strong opinions on this?
```
rpdelaney: AFAIK https authentication requires use of a password so ssh is (1) for sure most convenient and (2) _I think_ also more secure? The strength of our opinion would swing on the second point actually being true though.

developing/vcs/README.md
```diff
@@ -2,7 +2,321 @@
...
+change after updating the dependency.
+
+Cons:
+
+* Riskier since it requires a force push
+* Rewrites history
+
+Ways to mitigate cons:
+
+* Protect the master branch from force pushes
```
rpdelaney: We should consider recommending this be enabled no matter what else.

developing/vcs/README.md
```diff
@@ -2,7 +2,321 @@
...
+* Agree on a process where once a PR is under active review, no force pushes
+until the PR is approved and ready to be merged.
+* Original history can be viewed with `git reflog`
```
rpdelaney: We should only promise that the reflog will be available in the local repo. If you clone a new copy or run `git gc` you may lose  refs.

developing/vcs/README.md
macrael: We should recommend that the master branch be protected from _any_ pushes. The only way something goes to master is through CI

monfresh: For sure. See the section above under `GitHub repo settings` where I made this recommendation. GitHub labels it as protection from force pushes, but I'm pretty sure the setting `Require pull request reviews before merging` protects it from any kind of merge as well.

developing/vcs/README.md
```diff
@@ -2,7 +2,321 @@
...
+* [React](https://github.com/facebook/react)
+* [Bootstrap](https://github.com/twbs/bootstrap/blob/master/.github/CONTRIBUTING.md#pull-requests)
+* [Electron](https://electronjs.org/docs/development/pull-requests#step-6-rebase)
+* [Node](https://github.com/nodejs/node/blob/master/doc/guides/contributing/pull-requests.md#step-5-rebase)
+
+#### Examples of popular repos that enforce merge
+
+I'm trying to avoid confirmation bias, but I haven't found any yet. If you know
+of any, please add them. I found some that don't enforce any specific workflow,
+and most of them say they will squash your commits via the GitHub interface.
```
rpdelaney: Linux uses merge: [Linus on Merge vs Rebase](https://web.archive.org/web/20181107065817/http://thread.gmane.org/gmane.comp.video.dri.devel/34739/focus=34744), [kernel.org contributing docs](https://www.kernel.org/doc/html/latest/maintainer/rebasing-and-merging.html)

developing/vcs/README.md
```diff
@@ -2,7 +2,321 @@
...
+
+For each meaningful commit, write a
+[good commit message](https://chris.beams.io/posts/git-commit/) following these
+seven rules:
+
+1. Separate subject from body with a blank line
+2. Limit the subject line to 50 characters
+3. Capitalize the subject line
+4. Do not end the subject line with a period
+5. Use the imperative mood in the subject line
```
rpdelaney: We may want to link to something that explains what imperative mood is.

developing/vcs/README.md
```diff
@@ -2,7 +2,321 @@
...
+Pros:
+
+* Cleaner, more meaningful, easier to understand commit history
+* Easier to keep only commits that pass tests, which makes debugging
+easier with tools like `git bisect`.
+* Safer and easier to handle conflicts when commits are squashed
+* Easier to revert bad commits
+* Easier to change commit messages in a series of commits. We value great commit
+messages, so it should be easy to improve any message in a series of commits.
+* Allows reordering commits for more meaningful grouping. For example, you might
```
rpdelaney: I think the most idiomatic way to achieve this is aggressive local branching. But as discussed offline, that's not a universal practice.

developing/vcs/README.md
monfresh: Hmm. The second link doesn't say that they specifically prefer merge. They talk about the benefits of both. On GitHub, they point people to this document, which specifically ask people to either split their work into separate logical patches, or to squash commits using interactive rebase: https://www.kernel.org/doc/html/latest/process/submitting-patches.html

rpdelaney: This is that thing again where we are fighting the battle on the wrong field. The conflict isn't really about merging and rebasing. The contention, I think, is force pushing in a way that destroys published history, including (potentially) others' work. I don't think anybody objects to rebasing work that hasn't been published anywhere.

macrael: I definitely don't object to rebasing unpublished work. I use rebase from time to time, but I do think that it's a pretty sharp edged tool to be using as part of our recommended default flow, recommended to junior engineers who have a lot more important thinks to be worrying about than the intricacies of git.

rpdelaney: Remember the context is finding projects that use a merge workflow and eschew rebasing. I think Linux is the ultimate case of that, and should be mentioned if we will have a list.

From the kernel.org docs: 

> History that has been exposed to the world beyond your private system should usually not be changed...If work is in need of rebasing, that is usually a sign that it is not yet ready to be committed to a public repository.

On my reading, that uses a much more contrite tone but basically paraphrases and condenses a big part of Linus' email. An exception is mentioned but it's strictly confined to when git is being used to distribute work in progress, _not_ for others to base their own work on (since destroying published history is bad).

Here's the clear statement of intent, declaring allegiance to merge:

> Many projects require that branches in pull requests be based on the current trunk so that no merge commits appear in the history. The kernel is not such a project; any rebasing of branches to avoid merges will, most likely, lead to trouble.

The submitting patches doc where `rebase -i` is encouraged seems to be about creating pull requests that are likely to be accepted, but not rewriting published history.

monfresh: I'm realizing now that we might be talking about different things. There are two different concepts: one is rebasing in order to tidy up a PR *before* it is merged, and the other is how the PR ends up being merged into `master`. They are not the same thing. The way I read the second quote above deals with the second concept: how a branch/PR gets incorporated into the trunk (such as the `master` branch). GitHub provides 3 different ways to do that: merge commit, squash, and rebase. 

When the quote above refers to no merge commits appearing in the history, I understand that to be the equivalent of using the rebase button on GitHub, which does not add a merge commit (the squash button as well I believe does not add a merge commit).

The workflow I am proposing is that we always use the merge option on GitHub to merge a PR (which is what I had all along in the `GitHub repo settings` section), but if the PR has commits with poor messages or that are just fixing linting errors, or that cause tests to fail, those should be cleaned up first.

When I started that list of popular repos, I did not frame it correctly. I was only looking at their policy for preparing PRs, not how the PR is merged. I think that most people agree that it's a good idea to clean up PRs before they get merged, including Linux.

I will create a more specific poll to see what people think about cleaning up PRs and which buttons they prefer on GitHub.

rpdelaney: > I'm realizing now that we might be talking about different things. There are two different concepts: one is rebasing in order to tidy up a PR before it is merged, and the other is how the PR ends up being merged into master.

Yes. As far as I'm concerned this is about `push --force[-with-lease]`. 

>  I think that most people agree that it's a good idea to clean up PRs before they get merged, including Linux.

They do. In fact, they believe in it so much, they require that the PR should be cleaned up before it is even sent! But once it's sent, the history is published and should not be changed.

> History that has been exposed to the world beyond your private system should usually not be changed...If work is in need of rebasing, that is usually a sign that it is not yet ready to be committed to a public repository.

I think that means Linux & kernel.org would go in this part of the list.

macrael:
Here's a sketch of my git workflow:

1. make a new branch
2. write code
3. make commits when they make sense to me
    * Usually it's when a part of the feature I'm working on works
    * Sometimes it's when I need to work on something else
    * sometimes if it's just to work on something else, I'll reset that and keep working so my commits end being better scoped
    * usually I try and make sure tests are passing before I commit something
4. use rebase to clean up commits if need be
    * if I forget a file, or mispell a message I'll --amend
    * if there's a commit that doesn't really make sense on its own anymore I'll rebase it (rare)
5. incorperate new changes from develop: 
    * `git fetch`
    * `git merge origin/develop`
    * I always merge, and then don't rebase anything that came before the merge again
6. GOTO 2
7. Push, create a PR
8. GOTO 2
9. Merge the PR into develop

for whatever that's worth. I do try and make my commits make sense and pass tests, and I don't like to rewrite history once someone else's work is involved. 

Truly, I think the main difference between the two styles are 
1. I always use merge to bring in new changes from other folks, and 
2. I don't rebase everything into a single commit before merging
I think these are both reasonable practices so I'm fine with us continuing to allow both.

developing/vcs/README.md
monfresh: I try to follow Charlie Munger's advice about [fully understanding the other side's argument](https://fs.blog/2013/04/the-work-required-to-have-an-opinion/), but I don't think I do yet. Here is how I understand what you said. Please let me know where I misunderstood.

In the Linux example:
- You are equating submitting a patch with it being published, regardless of whether or not it is accepted. In other words, if someone submits a patch and the reviewer asks for changes, and if the submitter then creates several new commits locally to address the feedback, they can no longer rebase those commits to tidy them up before resubmitting.
- You are equating submitting a patch to Linux to submitting a PR to a repo in GitHub.

Is that what you are saying?

If so, that's not how I understand it, and I'd like to know where I might be wrong. The way I read the Linux contribution instructions [here](https://github.com/torvalds/linux/pull/415#issuecomment-301877514) is that a patch is sent via `git send-email`, and if it requires changes, another revision of the patch is sent using the version and annotate flags, such as `git send-email -v2 --annotate`, and the contributor describes what changed between versions 1 and 2.

In the GitHub PR above, the Linux maintainer tells the contributor that it's [fine to amend their commit](https://github.com/torvalds/linux/pull/415#issuecomment-302276507). The 2 patches are documented here:

[PATCH 1](https://marc.info/?l=linux-mm&m=149502833306198&w=2)
[PATCH v2](https://marc.info/?l=linux-mm&m=149504876613136&w=2)

The second patch doesn't include the original commit. It is an amended version of the first patch.

To me, that is the equivalent of using force push once a PR is approved, since that is the only way of updating the PR in order to do a final cleanup after addressing changes requested during PR review.

While a Linux patch is still being reviewed via email, I don't consider it to be published publicly. Only once it is merged into the "subsystem tree" would I consider it "exposed to the world".

In the world of GitHub, I do consider a PR to be published publicly, and I fully understand the risk of force pushing to a GitHub PR that other folks might be actively working on. What isn't yet clear to me is why folks think it's also risky to force push even if it's only done once at the end of a PR's life, when no one else will be touching that branch anymore.

To me, there isn't much difference between 1) pairing with someone locally and pushing the final PR and 2) pushing to GitHub first to have someone review the code remotely, making changes, and then force pushing to update the final commit message to document any important points that came up during the review.

The end goal should be to have a detailed commit message that also captures any relevant comments that might have been made during review. As I mentioned in Slack, I don't think it's a good idea to rely on GitHub as our canonical source of documentation. If there is relevant information in the comments of a PR, it should be summarized and captured in the final commit message. One shouldn't have to visit GitHub and try to read through all the comments, some or most of which might not be valuable, in order to get the big picture. I realize that the PR's main body can be updated on GitHub to summarize everything, but I don't think one should have to visit a website to understand a commit. If I want to look up a commit locally, I want to be able to stay in my Terminal or editor and not have to switch to a browser.

If we all agree that it's a good idea to document the PR in the commit message as opposed to on GitHub, and that it's OK to force push when it's necessary to update the final commit message, then I think we have a middle ground:

- those who prefer rebase can use it consistently, both to incorporate changes from the main branch, and to edit commits
- those who prefer merge can use it consistently, both to incorporate changes from the main branch, and to edit the final commit message using `git merge --squash`

In both cases, a force push is required at the end, and in both cases, we will use the "Merge" button on GitHub to incorporate the PR into the main branch. How does that sound?

developing/vcs/README.md
gidjin: My pattern is to only squash things before I've shared the PR for review. After that, I avoid it for commits that have been shared and only use merge to update my branch. Before my branch is in PR review I rebase with master.

gidjin:
My flow tends to be

1. Make a new branch
2. Code
3. Make commits when they make sense
    - Things are working
    - I need to stop for the day
    - An idea is complete
    - I want to leave a note for my post weekend self
4. These commits are always pushed to my branch end of day in case my laptop is not available for some reason
5. To update from master I `git pull --rebase`
6. Once I am ready, or if I have too many messy commits I will rebase to clean them up for consumption by others
7. Once the feature is complete I push the final code, if it's not already there, create a PR that is ready for review.
8. I `git merge origin/master` and add commits to the top for any review changes

I avoid rebasing when others have started reviewing my code or if I'm working on a branch with someone else

developing/vcs/README.md
```diff
@@ -2,7 +2,321 @@
 
 ## Overview
 
-Source control has less to do with the vcs system you use and more to do with the stages code goes through on the journey from inside a developers head, tested and merged into the main line, deployed to production and thence in time to being identified as that accursed legacy system that is causing all the problems.
+Source control has less to do with the VCS system you use, and more to do with the stages code goes through on the journey from inside a developer's head, tested and merged into the main line, deployed to production, and thence in time to being identified as that accursed legacy system that is causing all the problems.
+
+## Repo maintenance and hygiene
+
+* Avoid including files in source control that are specific to your development machine or process.
+* Ignore files that are specific to your machine or process in your
```
ferlatte: Why? Any time we tell people they need to do something, we need to explain very clearly why this is important.

developing/vcs/README.md
```diff
@@ -2,7 +2,321 @@
...
+`~/.gitignore_global` as opposed to your repo's `.gitignore`.
+* Delete local and remote feature branches after merging. This can be automated
```
ferlatte: Again: why? Also, if this *can* be automated, then we should either fully automate it, explain why we should do this manual toil, or decide that it doesn’t matter.

developing/vcs/README.md
```diff
@@ -2,7 +2,321 @@
...
+Pull requests should be small and focused enough to be reviewable in under 15,
+or ideally under 5, minutes.
+
+## Git workflow
+
+### Fork vs clone
+
+If it's a repo you have write and push access to, clone it. Otherwise, fork it.
+
+### Cloning with SSH vs HTTPS
```
ferlatte: I have strong opinions. Use https for read-only repos (ie, where you’re only ever going to clone and pull) and use ssh for everything else. I believe the why is because ssh supports certificates on security keys, and https does not; therefor, we should *not* use https for anything authenticated. However, if you’re just pulling a public repo, https is way more convenient since you don’t need to authenticate anything (unlike ssh for read-only repos).

developing/vcs/README.md
```diff
@@ -2,7 +2,321 @@
...
+
+Do we have strong opinions on this?
+
+### Merge vs rebase
+
+#### TL;DR
+
+If you care about keeping a clean commit history, where each commit has a great
```
ferlatte: So: do we care? What are the trade offs of caring?

I see advantages; however, if we’re investing a ton of effort to maintain a history that nobody ever looks at, I’m not sure it’s worth the time to maintain. So: are we committed to an engineering practice that uses clean history, teaching everyone how to use git bisect correct, and demonstrating the value of this practice? Right now, I see a lot of belief and not a lot of facts.

developing/vcs/README.md
```diff
@@ -2,7 +2,321 @@
...
+* Safer since you don't have to force push
+* History does not change, and includes the good, bad, and ugly
+
+Cons:
+
+* Noisier history with commits that don't add any value and that
+can cause confusion when reviewing the log
+* Not all commits might pass tests, making debugging hard
+* Cannot use interactive rebase to squash and/or reorder commits
+* Increases risk of losing code when handling conflicts due to working with
```
ferlatte: Are we sure about this one? Aren’t conflicts about the current state of the source and destination branches and don’t involve the history at all?

developing/vcs/README.md
```diff
@@ -2,7 +2,321 @@
...
+have a PR with several commits that each update a dependency. If the tests fail
+due to a dependency in the middle, you'll want to fix that in a separate commit,
+then rebase interactively to move the fix under the dependency commit and
+combine it with `fixup`, then edit the commit message to explain what had to
+change after updating the dependency.
+
+Cons:
+
+* Riskier since it requires a force push
+* Rewrites history
```
ferlatte: History re-writing never seemed to be a thing that actually mattered in terms of value generated by this process; however, force push is a problem.

developing/vcs/README.md
ferlatte: @monfresh I interpret the PATCH and PATCHv2 as... not like rebase and force push; conceptually, if you add a commit (PATCHv2), you’re not rebasing, you’re pushing a new commit into your branch.

I see what appears to be a consistent misunderstanding of the concerns around rebase: the problem isn’t “is this safe when everything goes right”, it’s “what happens when this doesn’t go right”. This is the fundamental concern with git workflows that require force push anywhere in the flow; at some point, someone’s going to typo a branch and have a bad day.

I’m actually ok with us accepting that risk if we think the trade off of a clean history (which, TBH, I think is over-valued, however, that’s a belief, not a fact). I am *not* ok with us having two ways to do the basic, every day, 100% needed operation. That seems ridiculous.

So: here’s what I want: do the risk analysis, with a focus on correctness and safety, in a way that (in theory) should convince this merry band of *engineers*  that it’s better to have one way to do this that we train and socialize, even if it’s not the way we currently believe to be best.

developing/vcs/README.md
```diff
@@ -2,7 +2,321 @@
...
+Clone the repo locally (example below uses SSH):
+
+```
+git clone git@github.com:trussworks/Engineering-Playbook.git
+cd Engineering-Playbook
+```
+
+Create a local feature branch based off master. Start the branch name with your
+initials. For example: `mb-git-worfklow`. If the branch is related to a Pivotal
+story, add the story's ID to the end of the branch: `mb-git-workflow-166068559`
```
ferlatte: Why the ceremony around initials and story ID?

developing/vcs/README.md
```diff
@@ -2,7 +2,321 @@
...
+channel, and/or via the GitHub site.
+
+Once you push to GitHub, only make changes via new commits, as opposed to
+amending or rebasing, to make it easier for reviewers to see what has changed
+since they last provided feedback. If you need to make a minor change to an
+earlier commit, take advantage of the `--fixup` flag, coupled with the
+`--autosquash` flag for interactive rebase, which you should enable globally:
+
+```
+git config --global rebase.autosquash true
```
ferlatte: Again: explain why, and explain (if possible, this is git after all) what this is doing.

developing/vcs/README.md
rpdelaney: Having the story id in the branch allows some convenient integration with the tracker; your branch (and any PRs sent from it) will show up in the story automagically.

Initials I think are meant to mark branch ownership... imho git itself is opinionated in that branches are not owned by anybody. But I do this usually since folks seem to like it and not much (anything?) is lost

monfresh: It also helps to quickly identify who to ping if there are stale branches in GitHub.

brainsik: GitHub shows branch authors in the UI, so initials aren't needed anymore for cleanup purposes. Since branches can be collaborative, I prefer having a good name without any ownership marker.

image

developing/vcs/README.md
monfresh: This was poorly worded by me and will be removed. See discussion above started by MacRae. Also, most of this will be rewritten. It was a first draft to start the discussion.

monfresh:
@rpdelaney This is ready for another review please.
@ferlatte I added some reasons for ignoring files, deleting merged branches, and small PRs. What do you think?

We are removing the contentious bits for now until we complete the risk analysis.

Latest commit is here: https://github.com/trussworks/Engineering-Playbook/pull/46/commits/3a107cd47490b053503daa60b19871b1f959c0cc

developing/vcs/README.md
ferlatte: Which one is preferable?

rpdelaney: Exposing my thinking for sanity checking what I believe these commands are doing:

* `fetch --prune` deletes local branches that are no longer present on the tracked remote.
* `fresh` (as defined here) looks like it is a bit more aggressive: it pulls to master and then deletes any local branches that are reachable from (have been merged to) master (excluding master itself), regardless of whether they have been deleted on the remote.

That might mean that a branch that has been deleted on the remote without being merged to master would remain in the local repo after `fresh`, but I haven't tested this.

@monfresh Am I in the neighborhood?

Which one is preferable, I'm not sure. We are recommending automating deleting merged branches, but `fresh` would get rid of them locally if that's not the case.

rpdelaney: @monfresh Doesn't `fresh` delete integration branches if you happen to run it straight after integration is merged to master?

monfresh: Are you saying that running `git fetch --prune` by itself should delete local branches that have been deleted from the remote? It's not working for me. When I run it, I see something like this:
```
 - [deleted]         (none)     -> origin/mb-squash-merge
```
but it doesn't actually delete my local branch. What am I missing?

Also, ideally, local branches would be deleted even if the branch hasn't yet been deleted from the remote. As long as it is merged, we should be able to delete the local branch.

rpdelaney: > Are you saying that running git fetch --prune by itself should delete local branches that have been deleted from the remote?

That is what I was saying, but I tried it again and I was wrong. From `git help fetch`:

>        -p, --prune
>           Before fetching, remove any remote-tracking references that no longer exist on the remote

So `--prune` removes the *tracking refererence* (what shows up in `git branch -v as something like `remotes/origin/branchname`) -- probably a good thing, but it does not delete the local branch at all. That still requires `branch -d`. 

I suppose an integration branch deleted by `fresh` could be retrieved with `fetch origin/integration`. So even in the worst case this shouldn't actually destroy any work.

PR #47, opened 2019-10-02
update lint install

nelztruss:
Install for `markdownlint` is via `npm` as per https://github.com/igorshubovych/markdownlint-cli (rather than via `brew`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,5 @@ One question we might ask is whether or not this repo should be public? My incli
 ### Install Dependencies(Mac OS)
 
     brew update
-    brew install npm pre-commit
-    npm install -g markdownlint-cli
+    brew install pre-commit
     pre-commit install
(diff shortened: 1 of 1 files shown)


brainsik:
You don't need to install _any_ version of markdownlint. This should be handled by `pre-commit` itself. In general, that's how `pre-commit` works. There are some exceptions where you do need to install an external binary (e.g., `terraform` and `shellcheck`) in order for the hook to do it's thing, but that's not the general case.

nelztruss:
Changed the PR to remove the errant instruction:
```
$ brew install markdownlint
Error: No available formula with the name "markdownlint"
```

monfresh:
Can we squash the commits before merging, please, so it's clearer what the issue was?

nelztruss:
Slack discussion: https://trussworks.slack.com/archives/CDJ9JAD26/p1570054331253700

PR #53, opened 2019-12-11
Yubikey guide

travelar:
Porting the YubiKey configuration guide to GitHub for longer term maintenance. 

@dynamike I moved the SSO stuff into a separate page as that will probably grow and it seemed like a better place.


travelar:
> yassssssssssss
> 
> only thing I would change is to add a note about who/where/how to ask questions. eg, should I open a pull request to a FAQ doc? DM you on slack? tweet you?

Was this not enough?  
If you have problems or questions during the configuration process, check in with #infrasec-chat on the Truss Slack.

PR #55, opened 2020-01-02
Adds guidance around semver

esacteksab:


esacteksab:
> I assume you used this convo: https://trussworks.slack.com/archives/CBRSZT93N/p1573251590089000 to inform.

Yes. It was referenced in the Pivotal ticket.

infra/tf/README.md
```diff
@@ -9,6 +9,13 @@ Terraform is our tool of choice for automating our 'cloud infrastructure'. In pa
 * Something about Truss Terraform Modules
 * How to layout/structure a Terraform Project
 
+### Semantic Versioning
+
+* When publishing a module for the first time, if you're uncertain what version to use, use `v.1.0.0`
```
rpdelaney:
```suggestion
* When publishing a module for the first time, if you're uncertain what version to use, use `v1.0.0`
```

jsclarridge:
This looks good, but I'm wondering if there are other topics to cover or clarify:
* Adding an input variable, but otherwise non-breaking - should that be a minor version update? It seems to fall under "add functionality in a backwards compatible manner".
* Renaming/removing an input variable - a major update? seems to fall under "incompatible API changes".
* Are there cases where recreating a resource would warrant a major version increment?

esacteksab:
>     * Adding an input variable, but otherwise non-breaking - should that be a minor version update? It seems to fall under "add functionality in a backwards compatible manner".

Agreed on minor version update

> Renaming/removing an input variable - a major update? seems to fall under "incompatible API changes".

Feels odd to do a major update on a rename, if the functionality of the module doesn't change. I think I'd be inclined to do a minor update in this case. _Removing_ a variable, I'm more inclined to do a major, but still struggling with that v1.0 -> v2.0 leap. But my experience in software dev is shallow. 

Curious what the others think.

nelztruss:
> Feels odd to do a major update on a rename, if the functionality of the module doesn't change. I think I'd be inclined to do a minor update in this case. _Removing_ a variable, I'm more inclined to do a major, but still struggling with that v1.0 -> v2.0 leap. But my experience in software dev is shallow.
> 
> Curious what the others think.

If an external consumer of the tool was already using a variable, then _has to_ update their usage to keep their usage working, then it _should be_ a full version "breaking" change.

esacteksab:
@jsclarridge any other thoughts?

PR #57, opened 2020-01-08
Add note about bash keyboard shortcuts on macOS

monfresh:
The Bash manual that we link to says you can move the cursor forward
or backward between words using the meta key and f or b. This isn't
the case by default on macOS. To avoid confusion, I thought I'd
clarify what the default keyboard shortcut is on macOS, and show folks
where they can change key mappings if they want.


rpdelaney:
Does this work differently when using bash 5x with homebrew? This doc recommends doing that, so it is probably worthwhile to notice either way.

PR #62, opened 2020-01-16
Document Truss guidance on exploratory learning

ahobson:


developing/growth/README.md
```diff
@@ -0,0 +1,23 @@
...
+speeding up tests) and the client getting the benefit of learning that
+happened elsewhere. Allowing time for growth ensures consistent, high
+quality work which benefits everyone.
+
+Mentoring is an important part of growth.  Truss has [shared some
+techniques for
+mentoring](https://truss.works/blog/2016/8/12/theres-a-method-to-the-mentoring-advice-from-a-teacher-turned-developer).
+
+To ensure that growth happens, the time should be built into the
+estimate for a ticket. From time to time, it will be appropriate to
```
rpdelaney: Does this mean that we should tend to avoid learning that isn't in the service of an existing user-driven story or ticket, or that learning tasks like that should have their own ticket? This also might have implications re: billing (or does it?)

ahobson: I added a note about learning non client related things.

rpdelaney: Great, thank you!

PR #69, opened 2020-02-21
Add Python content to eng playbook

jim:
This is a distillation of discussions we've had and research I've done about how to best write Python in our projects.

### DIFF
--- a/developing/languages/README.md
+++ b/developing/languages/README.md
@@ -14,7 +14,7 @@ While Truss may have opinions on particular languages and strategic reasons for 
 * Rust
 * Kotlin
 * [Ruby](./ruby/README.md)
-* Python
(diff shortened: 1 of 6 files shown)


rpdelaney:
AHH I was drafting something like this :)

jim:
@rpdelaney I would love to see what you've come up with.

rpdelaney:
From the looks of it I did a very similar structural change like you did, and then added some more opinionated stuff that could (potentially) be controversial but I think is nonetheless correct :-) With that in mind I suggest the thing to do here is merge in your PR and I'll adapt my additions into a follow-on for separate discussion. What do you think?

jim:
@rpdelaney Sounds good. I converted this to a real PR so it can be reviewed.

PR #71, opened 2020-03-06
Adding GovCloud organization setup guide

cblkwell:
Wanted to add at least an early version of my notes on how to build out an organization in GovCloud; I think there still needs to be more done here, especially with Cloudtrail, but I wanted to get some of this in there.

### DIFF
--- a/infra/aws/govcloud.md
+++ b/infra/aws/govcloud.md
@@ -24,7 +24,7 @@ security and compliance standards needed to meet federal guidelines like
   arn:aws:iam::123456789000:user/testuser
   ```
 
-  a GovCloud arn will look like this:
+  a GovCloud ARN will look like this:
 
   ```text
   arn:aws-us-gov:iam::123456789000:user/testuser
@@ -69,7 +69,7 @@ to all GovCloud accounts.
 ### Creating Additional GovCloud Accounts
 
 To add more accounts to your GovCloud organization, you *cannot* simply
-use the Terraform `aws_organizations_account" resource. Using credentials
+use the Terraform `aws_organizations_account` resource. Using credentials
 for the master account in your commercial AWS Organization, use the
 following AWS CLI command. This will create two accounts; a dummy account
 in the commercial AWS Organization and a new corresponding GovCloud
--- a/infra/aws/govcloud.md
+++ b/infra/aws/govcloud.md
@@ -1,0 +1,199 @@
+# Working in GovCloud
+
+Many of our government clients will want to have at least their production
+infrastructure in [AWS GovCloud](https://aws.amazon.com/govcloud-us/), which
+is a set of AWS regions that are separate from commercial AWS regions, with
+a number of additional controls that allow them to conform to the higher
+security and compliance standards needed to meet federal guidelines like
+[FedRAMP](https://www.fedramp.gov/).
+
+## Notable Considerations With GovCloud
+
+* Every GovCloud account must be attached to a corresponding account in
+  commercial AWS (mostly for billing purposes). This has some impact on
+  how you build AWS Organizations in GovCloud (see below).
+* Commercial AWS and GovCloud resources cannot interact directly in any
+  way. This means that you cannot assume roles between the two or tie
+  a commercial resource to a GovCloud resource unless using some publicly
+  addressable method.
+* GovCloud resources use a different AWS partition; this is the second
+  part of an ARN. So while an ARN for a commercial resource might look
+  like this:
+
+  ```text
+  arn:aws:iam::123456789000:user/testuser
+  ```
+
+  a GovCloud arn will look like this:
+
+  ```text
+  arn:aws-us-gov:iam::123456789000:user/testuser
+  ```
+
+  This means that any Terraform modules will need to take this into
+  account; see the Terraform
+  [`aws_partition`](https://www.terraform.io/docs/providers/aws/d/partition.html)
+  data source for how to do this.
+* Many AWS services are not available in GovCloud; some are available
+  in one GovCloud region and not others. See the GovCloud
+  [Services in Scope](https://aws.amazon.com/compliance/services-in-scope/)
+  and the AWS [Region
+  Table](https://aws.amazon.com/about-aws/global-infrastructure/regional-product-services/)
+  for more information.
+* Cloudtrail is automatically added to all accounts in GovCloud by default.
+  This means that you will have to handle things differently when trying to
+  configure an organization's Cloudtrail.
+
+## Setting Up a GovCloud Organization
+
+### Create Your GovCloud Organization Master Account
+
+To set up a new GovCloud AWS Organization, you must log in as root to the
+master account of your commercial AWS Organization. Then go to the "My
+Account" screen with the pulldown in the upper right of the AWS dashboard,
+and scroll down; there will be a "Sign up for AWS GovCloud" button. When
+you click that, you will need to accept an appendix to the AWS Customer
+Agreement; once you do so, you will get an email with instructions on how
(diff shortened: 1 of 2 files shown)


infra/aws/govcloud.md
```diff
@@ -0,0 +1,199 @@
...
+  account; see the Terraform
+  [`aws_partition`](https://www.terraform.io/docs/providers/aws/d/partition.html)
+  data source for how to do this.
+* Many AWS services are not available in GovCloud; some are available
+  in one GovCloud region and not others. See the GovCloud
+  [Services in Scope](https://aws.amazon.com/compliance/services-in-scope/)
+  and the AWS [Region
+  Table](https://aws.amazon.com/about-aws/global-infrastructure/regional-product-services/)
+  for more information.
+* Cloudtrail is automatically added to all accounts in GovCloud by default.
```
dynamike: TIL

infra/aws/govcloud.md
```diff
@@ -0,0 +1,199 @@
...
+  This means that you will have to handle things differently when trying to
+  configure an organization's Cloudtrail.
+
+## Setting Up a GovCloud Organization
+
+### Create Your GovCloud Organization Master Account
+
+To set up a new GovCloud AWS Organization, you must log in as root to the
```
dynamike: Do you use the existing org-root AWS account for the rest of commercial or do create a new master commercial AWS account which will be org-root-gov or something?

infra/aws/govcloud.md
```diff
@@ -0,0 +1,199 @@
...
+To add more accounts to your GovCloud organization, you *cannot* simply
+use the Terraform `aws_organizations_account` resource. Using credentials
+for the master account in your commercial AWS Organization, use the
+following AWS CLI command. This will create two accounts; a dummy account
+in the commercial AWS Organization and a new corresponding GovCloud
+account, both with the same account alias.
+
+```console
+$ aws organizations create-gov-cloud-account \
+  --email spacecats-infra+spacecats-govcloud-id@example.com \
```
dynamike: fucking spacecats

infra/aws/govcloud.md
```diff
@@ -0,0 +1,199 @@
...
+$ terraform import aws_organizations_account.spacecats_govcloud_id 000987654321
+```
+
+### Handling GovCloud Dummy Accounts in AWS Commercial
+
+As stated above, when you use the AWS CLI to create your new GovCloud
+accounts, it will also create a dummy account with the same account name
+in your commercial AWS organization. In order to prevent misuse of these
+accounts (and any confusion, since they have the same account alias as
+your GovCloud account), we can lock these down with the `suspended` OU
```
dynamike: NICE

infra/aws/govcloud.md
chrisgilmerproj: I don't have an Organization set up for my AWS account (or maybe the root account is). So does that change things? Or would you suggest starting with AWS Org setup before starting GovCloud work?

infra/aws/govcloud.md
```diff
@@ -0,0 +1,199 @@
...
+To set up a new GovCloud AWS Organization, you must log in as root to the
+master account of your commercial AWS Organization. Then go to the "My
+Account" screen with the pulldown in the upper right of the AWS dashboard,
+and scroll down; there will be a "Sign up for AWS GovCloud" button. When
+you click that, you will need to accept an appendix to the AWS Customer
+Agreement; once you do so, you will get an email with instructions on how
+to log in to your GovCloud account.
+
+Once you create the account, you can create a temporary IAM user with full
+credentials you can use to bootstrap Terraform in the account as you would
```
chrisgilmerproj: A link to our bootstrap recommendations would be great.

infra/aws/govcloud.md
chrisgilmerproj: is this just any email from `truss.works`? And is the account name just the AWS Account ID?

infra/aws/govcloud.md
```diff
@@ -0,0 +1,199 @@
...
+
+In that output, `AccountId` is the account number for the dummy account,
+and the `GovCloudAccountId` is the account number for your new GovCloud
+account.
+
+Now that you have that, you can assume the `OrganizationAccountAccessRole`
+in the new account from an account with sufficient privileges in the
+*GovCloud* organization master account. *Note that this is the case even
+though at this point the GovCloud organization does not show up if you
+look at the organization membership in the console.*
```
chrisgilmerproj: An example of assuming roles would be nice. Are you doing it from the CLI or via the AWS config file?

infra/aws/govcloud.md
```diff
@@ -0,0 +1,199 @@
...
+we usually create in the commercial AWS organization.
+
+To do this, import the account's Terraform resource into your commercial
+AWS organization by using the same sort of procedure described above to
+import the new GovCloud account into the GovCloud organization, but
+replace the `parent_id` parameter with the Terraform resource describing
+your `suspended` OU -- usually something like
+`aws_organizations_organizational_unit.suspended.id`. This will make it
+impossible to use that account for anything else, and has no effect on
+the corresponding GovCloud account.
```
chrisgilmerproj: Can you also link the terraform docs for import of this resource?

infra/aws/govcloud.md
cblkwell: @dynamike I used the existing org-root account when I did this; doing that will keep all your billing tied to that same account, which makes things simpler, and there's no real reason to create a whole new account just to spin up your GovCloud organization.

@chrisgilmerproj I would recommend setting up an organization for your commercial account(s) first before starting off with GovCloud. That allows you to use the `aws organizations` commands to create the GovCloud accounts (since you can't do that without an organization, I assume). You could always create a new account (milmove-org-root or whatever), create your organization there, and then move your existing account into the new organization. That will let you put off your migration of AWS resources into an organization-centric arrangement in commercial, but still be able to tie your GovCloud account to it and do GovCloud in an organization-centric fashion.

infra/aws/govcloud.md
cblkwell: @chrisgilmerproj The email is the email address you want to have AWS send notifications about the account to. So in this case, `spacecats-infra@example.com` is the email alias for the infra team on the Spacecats project (with a suffix for that particular GovCloud account), and `spacecats-govcloud-id` is just the account alias for the new account.

infra/aws/govcloud.md
cblkwell: I usually just add a profile for it in `.aws/config` -- that tends to be easier, since `aws-vault` will pull it for all our `aws` and `terraform` commands.

infra/aws/govcloud.md
cblkwell: Yeah, I was pretty happy when I figured out this was fine.

infra/aws/govcloud.md
cblkwell: What do you mean? Do you just mean this page: https://www.terraform.io/docs/import/usage.html or something else? I gave an example of how to do this in the previous section, where I showed how to do it for the GovCloud account.

infra/aws/govcloud.md
chrisgilmerproj: Aha! I'd put a note about that somewhere here. Also maybe point out that using the `+` is a good way to identify senders.

infra/aws/govcloud.md
chrisgilmerproj: Not the `terraform import` docs but the page for the terraform resource. Like this one: https://www.terraform.io/docs/providers/aws/r/organizations_organizational_unit.html#import

PR #73, opened 2020-03-09
add 2 gotchas worth memorializing

kilbergr:
I would like to write a more in-depth explanation of this tool and how we use it. Until then, I'd like to record some gotchas that we've run into. This is the place to do it!

### DIFF
--- a/infra/tf/atlantis.md
+++ b/infra/tf/atlantis.md
@@ -1,8 +1,8 @@
-## Atlantis Truss
+# Atlantis Truss
 
 We're beginning to use Atlantis throughout projects at Truss. You can explore the Atlantis setup in Truss' commercial accounts for the prod tiers [here](https://github.com/trussworks/legendary-waddle) and the dev tiers [here](https://github.com/trussworks/legendary-waddle-dev). There's more to come, but for now here are some helpful things to know.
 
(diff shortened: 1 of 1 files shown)


infra/tf/atlantis.md
```diff
@@ -0,0 +1,8 @@
+# Atlantis Truss
+
+We're beginning to use Atlantis throughout projects at Truss. You can explore the Atlantis setup in Truss' commercial accounts for the prod tiers [here](https://github.com/trussworks/legendary-waddle) and the dev tiers [here](https://github.com/trussworks/legendary-waddle-dev). There's more to come, but for now here are some helpful things to know.
+
+## Tips and Gotchas
+
+* When resetting a GitHub personal token in Parameter Store, you will have to redeploy the Atlantis instance. Otherwise, the new credential will not be updated in the instance.
```
cblkwell: I suspect you actually just need to restart the atlantis task in ECS, technically.

kilbergr: In fact, that is not correct! I tried that multiple times and it wasn't doing the trick--it's because of the daisy-chaining between the module-created ssm param that populates after pulling down your secret from your var and your var.

PR #80, opened 2020-03-25
Add some goreleaser documentation

eeeady:
Bad branch name.

This adds goreleaser documentation with lotsa examples. My eyes are fried. I need 👀 on this. ❤️


esacteksab:
Reading through this, I'm left wondering if there wouldn't be value in a Github repo that's a template much like the Terraform Module Template https://github.com/trussworks/terraform-module-template

eeeady:
There is... I put a placeholder for that in the backlog... Depends on if we wind up using this pattern I think. I can spend the time doing that though!

developing/command-line-tools/HOW2GORELEASER.md
```diff
@@ -0,0 +1,357 @@
...
+Create a `Dockerfile` at the root of your repo and commit it to git.
+
+### Example Dockerfile
+
+This example is a very basic Dockerfile that works with goreleaser with a tool/binary name of `binaryname`.
+
+Goreleaser will use the binary it determines is correct to copy into the container.
+
+```docker
+FROM alpine:3
```
dynamike: Would it make sense to limit this to something like scratch or one of the distroless containers?

eeeady: Maybe? probably? I'm not familiar with what are the tradeoffs.

PR #83, opened 2020-04-13
Add tldr as recommended command line tool

ralren:
What the title says.

### DIFF
--- a/developing/command-line-tools/README.md
+++ b/developing/command-line-tools/README.md
@@ -7,12 +7,12 @@ developer happiness.
 
 You may also be interested in these other pages from the Truss Engineering Playbook:
 
-* [docker](https://github.com/trussworks/Engineering-Playbook/tree/master/developing/docker)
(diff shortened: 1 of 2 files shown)


nelztruss:
@ralren Did a linter change all the "*" to "-"? Was that intentional?

ralren:
@nelztruss not intentional. Must have been a linter.

nelztruss:
@ralren So, I'm not sure how process-y we should be in this repo. Usually, it's considered to be a good idea to break up substantive changes from formatting changes into different commits. (Also, it'd probably be a good idea to figure out why the linter all of a sudden changed its mind on "-" vs "*" for bullets, was it because of a newly release version or something?)
I'll give you the thumbs-up so you can commit if you don't feel that these suggestions are germane for this project.

ralren:
@nelztruss, I'll at least spend fifteen minutes debugging what's going on with the linter.

PR #89, opened 2020-04-24
Node.js release guidance

tinyels:
This comes up a lot on slack, so trying to consolidate advice here. This is fairly opinionated, so welcome comments.

### DIFF
--- a/web/frontend/README.md
+++ b/web/frontend/README.md
@@ -59,9 +59,11 @@ We generally adhere to the [Airbnb JavaScript Style Guide](https://github.com/ai
 * [ESLint](https://eslint.org).
 
 #### Selecting a release of Node.js
+
 When starting a new project, you will want to select an Active LTS release of Node.js and make plans to move to the next LTS release after it becomes active and before the selected release goes out of maintenance. You could choose a "current" release but there is a good chance that the libraries you want to use have not been tested until it becomes active. Odd-numbered releases are short-lived, so only choose one intentionally. (For instance, if some library you want to use doesn't work in an LTS release.) Please refer to this [release schedule](https://nodejs.org/en/about/releases/) to make your selection. 
 
-Once you have done this, you may want to enforce it by configuring the `engines` your projects package.json. For instance, if you have selected 12.x 
+Once you have done this, you may want to enforce it by configuring the `engines` your projects package.json. For instance, if you have selected 12.x, you might want to add this section:
+
 ```
 # package.json
 "engines": {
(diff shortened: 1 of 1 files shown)


web/frontend/README.md
```diff
@@ -58,6 +58,28 @@ We generally adhere to the [Airbnb JavaScript Style Guide](https://github.com/ai
...
+```
+
+If your team is managing tooliing with brew, you will want them to install and pin the specific version:
+
+```
+brew install node@12
+brew pin node@12
+```
+
+(Note that nvm isn't supported if it is installed with brew.)
```
haworku, line 81: 🤔 I always thought you should manage node versions with `nvm` over homebrew.  Definitely something that was "told" to me -  not something I am opinionated on personally.   Does Truss suggest installing Node with homebrew or is that the general pattern here?

web/frontend/README.md
```diff
@@ -58,6 +58,28 @@ We generally adhere to the [Airbnb JavaScript Style Guide](https://github.com/ai
...
+Once you have done this, you may want to enforce it by configuring the `engines` your projects package.json. For instance, if you have selected 12.x, you might want to add this section:
+
+```
+# package.json
+"engines": {
+  "node": "^12"
+}
+```
+
+If your team is managing tooliing with brew, you will want them to install and pin the specific version:
```
gidjin, line 74: minor typo `tooling`

web/frontend/README.md
gidjin: At previous job we used rbenv for ruby and nvm for node. Pinning homebrew to a specific release means you only have that one option for all projects using that tool, which can be an issue. I know MilMove went the route of using brew but I personally would prefer nvm or, even better IMO, having a Dockerfile that pulled a specific version of node for the project to run in a container. The latter has the advantage of keeping the exact same version of node locally as is run in your build and deploy, assuming docker is used in your build pipeline and deploy process.

I'm wondering if we should add a couple more sections here, for example: _if you are using `nvm` here is how you choose your version_ and _if you are using docker here is an example Dockerfile._ Unless we do have a specific recommended one.

tinyels: I could take this section out (or leave placeholders for other people to add more options). This has been a gotcha on milmove and I wanted at least to address it. This PR is pretty much so I have a link I can send instead of stating this over and over again on slack.

monfresh: I'm not sure pinning with brew should even be a recommendation. I don't think there's anything wrong with certain tools coming from brew, and others from elsewhere. I would prefer that we standardize on one tool, like `nvm` (similarly to how we standardized on `chruby` for Ruby), and use a script that guarantees everyone is installing it in the same way, like this for example: https://github.com/18F/laptop/blob/master/mac#L220-L240. I use `nvm` for MilMove and I haven't had any issues.

The only issue I have personally seen with projects that use yarn and node is that if yarn is installed via homebrew, it also installs node, so you have to run `brew uninstall --ignore-dependencies --force node` after upgrading yarn, but that can easily be automated via a standardized script like the `laptop` one above.

I think it's much better to have one supported way to do things and have them defined in a script so that there's less to maintain and troubleshoot.

tinyels: I am tempted to remove the section on managing the version since it is distracting from what I really wanted to accomplish here.

PR #90, opened 2020-04-24
Write documentation around CircleCI orbs.

ralren:
What the title says.

### DIFF
--- a/infra/circleci/orbs.md
+++ b/infra/circleci/orbs.md
@@ -1,0 +1,209 @@
+# [Infra](../README.md) / Orbs
+
+## Overview
+
+As defined by CircleCI, orbs are "a reusable package of YAML configuration that condenses repeated pieces of config into a single line of code." You can think of them as libraries you can import into the CircleCI configuration.
+
+There is a [registry of CircleCI orbs](https://circleci.com/orbs/registry/) including those created and managed by CircleCI and their partners.
+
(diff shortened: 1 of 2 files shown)


infra/circleci/orbs.md
```diff
@@ -0,0 +1,209 @@
...
+        steps:
+            - run:
+                command: < the actual logic behind what your orb does>
+
+# the environment the job will run in
+executors:
+    < placeholder-executor >:
+        # can use a machine or some other environment
+        docker:
+            - image: < placeholder-image >
```
cblkwell, line 153: This is more of a mechanical question, but when I write a CircleCI config I specify an executor as well -- does the orb run in the executor for the orb config, or the executor for the config I write in my `config.yml`? Does an orb *need* an executor?

ralren: To your more general comment:
I was personally torn between a generic example and a more specific example. I definitely stand to benefit from a specific example myself but I thought that a compromise of a general example while pointing to the orb we do have published would be sufficient enough.

To the more specific question:
I don't believe an orb always needs an executor. 

If you're using an orb's job, you would be using the executor (the one in the orb config) specified in that job. Like in L159. 

If you're using an orb's command, the orb will use the executor specified in the config it was imported to. In this case, you're `config.yml`.

PR #94, opened 2020-04-30
Adding teardown guide

cblkwell:
This adds the teardown guide for infrastructure; I didn't go into a ton of specific detail because I didn't think it was that necessary (the people executing this should *probably* be the most senior infra engineers on the project, so I didn't want to make it too handholdy), but I think I got enough detail that the major hitching points should be covered.

### DIFF
--- a/infra/teardown.md
+++ b/infra/teardown.md
@@ -57,6 +57,14 @@ some guidelines when tearing down Terraform namespaces:
   from the console or using the CLI) and then close it from the My
   Account panel.
 
+## SSL Certificates
+
+For most projects, we'll hopefully be able to use AWS ACM certificates,
+and those will get torn down with our Terraform teardown above. However,
+if we've bought additional SSL certificates through another vendor, such
+as SSLMate, we should revoke those certificates and close that account
+as well.
+
 ## CircleCI
 
 Assuming you've used CircleCI for your CI/CD pipeline on the project,
--- a/infra/teardown.md
+++ b/infra/teardown.md
@@ -1,0 +1,96 @@
+# [Infra](./README.md) / Project Teardown Guide
+
+When a project reaches the end of its lifetime and the client doesn't
+want to continue it, we need to teardown all the infrastructure we've
+created for the project -- sometimes on relatively short notice. This
+guide is intended to provide a step-by-step procedure for tearing down
+infrastructure we've created for a project and highlight any potential
+sources of trouble.
+
+## Terraform and AWS
+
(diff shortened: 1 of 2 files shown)


esacteksab:
LGTM. Having done this, the process is painful. I'm wondering if there isn't a more direct route with something like the Nuker we use to clean up unused resources or use something like Cloud Custodian and come at the environment a bit sideways if the intent is to get rid of _everything_. 

Just thinking out loud here more than anything.

infra/teardown.md
```diff
@@ -0,0 +1,96 @@
...
+## CircleCI
+
+Assuming you've used CircleCI for your CI/CD pipeline on the project,
+you should be able to just terminate your CircleCI support plan as soon
+as you don't need to do anymore deploys or validate code. You can do this
+from the "Project Settings" page on the CircleCI dashboard.
+
+## GitHub
+
+Once you've torn down your AWS infrastructure and CircleCI, you can
```
sojeri, line 69: why do we recommend destroying the whole github org? we often use tf to control who has access, so wouldn't it be enough to remove all trussels from the org?

cblkwell:
> LGTM. Having done this, the process is painful. I'm wondering if there isn't a more direct route with something like the Nuker we use to clean up unused resources or use something like Cloud Custodian and come at the environment a bit sideways if the intent is to get rid of _everything_.
> 
> Just thinking out loud here more than anything.

There may indeed be a better way, but I wanted to write down what I did to tear down stuff for SABER to save someone else (perhaps my future self) from boning themselves like I did -- basically everything I call out specifically *not* to do, I did. :P

cblkwell:
> I might put something in here about certificates. Depending on the life of the project and the folks running it, perhaps certs were set up to auto-renew. Maybe a blurb about checking and disabling auto-renew settings on certificate providers.

That's a good call. I didn't set anything for SABER to auto renew so that was pretty easy, but I can see other folks not doing that.

infra/teardown.md
cblkwell: If we are paying for the organization, and the client doesn't want to take it over, it has to go somewhere. A GitHub organization isn't just a logical collection of people like a team, it's a paid product that owns repos and contains teams.

kilbergr: I don't think you want a dangling org with no purpose.

cblkwell:
> V thorough as usual. Sounds like you ran into some gotchas ;)

Unfortunately. :P

PR #96, opened 2020-05-01
Mdawn  add test metadata steps for circleci 172547059

mdawn:
# [Tracker story](https://www.pivotaltracker.com/story/show/172547059)

-  Adds section & updates config to use CircleCI's test metadata using [`go-junit-report`](https://github.com/jstemmer/go-junit-report).

Next PR: update [the template](https://github.com/trussworks/terraform-module-template)

Question: is there a markdown way to prettify make or is it just text?


infra/tf/terratest.md
```diff
@@ -133,6 +137,53 @@ Update the `rotate.yaml` file in [Legendary Waddle Dev](https://github.com/truss
...
+```
+
+Finally we update our `.circleci/config` by adding two steps - one to `go get` the package and another to access our shiny new executable:
+
+```yaml
+    - run:
+        name: Adding go binaries to $PATH
+        command: |
+          echo 'export PATH=${PATH}:~/go/bin' >> $BASH_ENV
+          source $BASH_ENV
```
cblkwell, line 183: I don't think this is actually needed...did you see this in CircleCI docs somewhere?

mdawn: Oh. I thought because I set an `environment` I needed to add it bc of this: https://circleci.com/docs/2.0/env-vars/#using-parameters-and-bash-environment

Maybe I'm misunderstanding?

cblkwell: Hrm. Well, I know we didn't need that on SABER but also I think this might be more specifically about changing the path -- which we *didn't* do on SABER. Since you *are* changing the path, I think you might have the right of things.

cblkwell: I should make it clear, what I was referring to was the `source $BASH_ENV` line. :)

PR #104, opened 2020-05-15
Adding documentation section

cblkwell:
This is to start a documentation section in the Engineering Playbook to help folks write better documentation -- a lot of this is just stuff I've learned over the course of the last X number of years since I'm not really a technical writer by training, so review by someone who *is* a technical writer by training is probably a good idea. The intended audience here is specifically engineers.

### DIFF
--- a/documentation/external-resources.md
+++ b/documentation/external-resources.md
@@ -9,6 +9,7 @@ especially interested in this topic.
 * [10 Tips For Making Your Documentation Crystal Clear](https://opensource.com/life/16/11/tips-for-clear-documentation), Ben Cotton
 * [A Primer on Documentation Content Strategy](https://increment.com/documentation/primer-on-documentation-content-strategy/), Stephanie Blotner
 * [What to Write](https://jacobian.org/writing/what-to-write/) and [Technical Style](https://jacobian.org/writing/technical-style/), Jacob Kaplan-Moss
+* [Yelp Production Engineering Documentation Style Guide](https://engineeringblog.yelp.com/2018/10/yelp-production-engineering-documentation-style-guide.html), Chastity Blackwell
 
 ## Talks
 
--- a/documentation/external-resources.md
+++ b/documentation/external-resources.md
@@ -1,0 +1,17 @@
+# [Documentation](./README.md) / External Resources
+
+This is a list of articles and talks that have influenced the advice in
+this section; you may find it useful to read or view these if you're
+especially interested in this topic.
+
+## Articles
+
+* [10 Tips For Making Your Documentation Crystal Clear](https://opensource.com/life/16/11/tips-for-clear-documentation), Ben Cotton
+* [A Primer on Documentation Content Strategy](https://increment.com/documentation/primer-on-documentation-content-strategy/), Stephanie Blotner
+* [What to Write](https://jacobian.org/writing/what-to-write/) and [Technical Style](https://jacobian.org/writing/technical-style/), Jacob Kaplan-Moss
+
+## Talks
+
+* [The 7 Deadly Sins of Documentation](https://www.usenix.org/conference/srecon18europe/presentation/blackwell), Chastity Blackwell
+* [Scalable Meatfrastructure](https://www.usenix.org/conference/lisa15/conference-program/presentation/goldfuss), Alice Goldfuss
+* [Traps and Cookies: A Mystery Package From Your Former Self](https://www.usenix.org/conference/lisa16/workshop-program/presentation/reilly), Tanya Reilly
--- a/documentation/intro-to-docs.md
+++ b/documentation/intro-to-docs.md
@@ -109,7 +109,8 @@ In addition, adding lots of graphics will make it harder to read the
 text. If people cannot follow from one paragraph to the next (seeing them
 both on the screen at the same time), this is especially difficult. If
 you need to use large graphics, use thumbnails and link to the larger
-version to prevent them from being too distracting.
+version to prevent them from being too distracting. Graphics can also
+make your docs less accessible, so captions and alt text are important.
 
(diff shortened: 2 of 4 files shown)


documentation/external-resources.md
```diff
@@ -0,0 +1,18 @@
...
+
+This is a list of articles and talks that have influenced the advice in
+this section; you may find it useful to read or view these if you're
+especially interested in this topic.
+
+## Articles
+
+* [10 Tips For Making Your Documentation Crystal Clear](https://opensource.com/life/16/11/tips-for-clear-documentation), Ben Cotton
+* [A Primer on Documentation Content Strategy](https://increment.com/documentation/primer-on-documentation-content-strategy/), Stephanie Blotner
+* [What to Write](https://jacobian.org/writing/what-to-write/) and [Technical Style](https://jacobian.org/writing/technical-style/), Jacob Kaplan-Moss
```
duncan-truss, line 11: ^He wrote the Django documentation I believe?  I know Django was also born out of a Newspaper CMS so that's not surprising they are writing experts.

documentation/intro-to-docs.md
```diff
@@ -0,0 +1,153 @@
...
+spending pages defining common terms they're already familiar with will
+probably cause them to lose patience and start to skim things over.
+
+### Avoid going overboard with graphics or video
+
+Often we'll be tempted to add lots of screenshots, diagrams, or other
+graphics to documentation to try and make things more clear. A well-placed
+graphic can help -- but too many will make the upkeep of your documentation
+too difficult. It is *much* more difficult to edit a graphic (and videos
+even more so) than to edit text, so use them only when they worth the
```
duncan-truss, line 105: they -> they're

documentation/intro-to-docs.md
```diff
@@ -0,0 +1,153 @@
...
+
+### Keep your audience in mind always
+
+The most important thing when writing documentation is to remember that
+you are writing this *for* someone. It may be new engineers, it may
+be people on-call, it may be stakeholders in the C-suite, but whoever they
+are, you intend for *someone* to be reading this at some point in the
+future.
+
+If you are keeping this audience in mind throughout the process of writing
```
duncan-truss, line 92: are -> are not

documentation/intro-to-docs.md
```diff
@@ -0,0 +1,153 @@
...
+your docs, then you are going to miss the mark. Writing a document intended
+for new engineers and assuming a ton of knowledge will severely limit its
+utility. Similarly, writing something for intended for project leads and
```
duncan-truss, line 95: extra "for" in there

documentation/external-resources.md
pkarman: Jacob will be proud to know he's a writing expert. Django was born down the street from me in Lawrence, for the newspaper we read every day.

cblkwell: Well, I can't speak for anyone else, but I found his articles pretty helpful at least. ;)

documentation/intro-to-docs.md
```diff
@@ -0,0 +1,153 @@
...
+future.
+
+If you aren't keeping this audience in mind throughout the process of writing
+your docs, then you are going to miss the mark. Writing a document intended
+for new engineers and assuming a ton of knowledge will severely limit its
+utility. Similarly, writing something intended for project leads and
+spending pages defining common terms they're already familiar with will
+probably cause them to lose patience and start to skim things over.
+
+### Avoid going overboard with graphics or video
```
duncan-truss, line 99: In the technical writing course there was a tip to write the caption first for any image you wanted to use and make sure that's what the image actually conveys when someone looks at it.

documentation/intro-to-docs.md
```diff
@@ -0,0 +1,153 @@
...
+version to prevent them from being too distracting.
+
+### Keep individual documents short -- but not too short
+
+There's nothing worse than trying to find the one paragraph you need from
+a 20 page document. In order to avoid this, we should keep individual
+documents short and focused. Generally, anything more than 4 "pages" (ie,
+you would need to scroll four full screens to see the whole thing) is going
+to feel "long" to most readers -- if you go over this length, be aware that
+a lot of people will resist reading this documentation in a single sitting.
```
duncan-truss, line 121: Yea I think it's worth calling out that many people will skim the documentation so lists and tables are preferred over paragraphs of description.

duncan-truss:
Some other points that I found useful but not sure if they have a place here:
- Preferring active voice over passive voice so the reader doesn't have to mentally convert and the subject is less ambiguous
- Preferring sample programs over snippets that may not get retested over time (debatable)
- Just like when you are writing code, comments in examples should reflect WHY over WHAT
- Conditionals should come before not after imperative commands so you don't have any example like "Delete X, if  you no longer need Y" and the user has already carried out the task.
- In addition to audience it might be helpful to outline the purpose or desired outcomes and keep rechecking that hypothesis.

cblkwell:
> Some other points that I found useful but not sure if they have a place here:
> 
> * Preferring active voice over passive voice so the reader doesn't have to mentally convert and the subject is less ambiguous
> * Preferring sample programs over snippets that may not get retested over time (debatable)
> * Just like when you are writing code, comments in examples should reflect WHY over WHAT
> * Conditionals should come before not after imperative commands so you don't have any example like "Delete X, if  you no longer need Y" and the user has already carried out the task.
> * In addition to audience it might be helpful to outline the purpose or desired outcomes and keep rechecking that hypothesis.

I've tried to add most of these -- I think some of them were already in the text (perhaps with different words -- "if you cannot adequately summarize what a reader is supposed to get from the document in a single sentence, it is probably too widely scoped" seems to echo that last sentiment, for instance). Thanks for suggesting these.

PR #105, opened 2020-05-18
Use main as default branch instead of master

nelztruss:


nelztruss:
Truss has already decided against this as per the following TDR: https://docs.google.com/document/d/1hzSTD5mMVsfF_oO_bmeVzQnKDzG_22YYzC5pobJRGeg/edit

PR #106, opened 2020-05-19
Incident response section start

cblkwell:
So, this is just a start for the incident response section, and I'm not sure I'm *entirely* happy with this, but I wanted to get something there and I figure we can build on it later.

The big thing I think we're missing here is something on incident retrospectives; I thought we had a document somewhere in the Google Drive about this that I was going to use, but I can't find it now. I also think the Etsy Debriefing Guide is a pretty good place to start, but I thought about trying to put together something a little shorter. I've never actually been through an incident retrospective here either, so I am not sure what our usual process is. As a result I decided to leave it out for now and revisit it later -- I did add a bunch of my favorite articles/talks/etc on the topic to the external resources page.

I also moved the on-call guidance document here, since this seemed like a better place than the development section.

### DIFF
--- a/incident-response/analysis.md
+++ b/incident-response/analysis.md
@@ -1,8 +1,8 @@
 # [Incident Response](./README.md) / Incident Analysis and Retrospectives
 
 Once an incident has ended, we want to perform an analysis and conduct a
-retrospective to figure out what we can learn from what happened. It is
-important that we remember that **the primary purpose of any retrospective
-is learning, *not* necessarily fixing.**
+blameless retrospective to figure out what we can learn from what happened.
+It is important that we remember that **the primary purpose of any
+retrospective is learning, *not* necessarily fixing.**
 
 [This is a stub and needs to be expanded.]
--- a/incident-response/analysis.md
+++ b/incident-response/analysis.md
@@ -1,0 +1,8 @@
+# [Incident Response](./README.md) / Incident Analysis and Retrospectives
+
+Once an incident has ended, we want to perform an analysis and conduct a
+retrospective to figure out what we can learn from what happened. It is
+important that we remember that **the primary purpose of any retrospective
+is learning, *not* necessarily fixing.**
+
+[This is a stub and needs to be expanded.]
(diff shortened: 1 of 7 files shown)


incident-response/overview.md
```diff
@@ -0,0 +1,198 @@
...
+### Communicator
+
+The communicator is responsible for handling external communication during
+the incident, both by keeping the rest of the engineering team informed
+and notifying stakeholders (such as clients) about the status of the
+incident. Similarly, if non-responders want to ask questions or notify
+the incident response team of issues, they should use the communicator as
+a conduit to the incident commander, not talk to the commander directly.
+
+During less severe incidents, the communicator role may be folded into
```
travelar, line 68: > During less severe incidents, the communicator role may be folded into
the scribe role.

I don't know that ever seen an actual communicator role in use. I typically see the IC handle most communications, especially with the customer.

incident-response/analysis.md
```diff
@@ -0,0 +1,8 @@
+# [Incident Response](./README.md) / Incident Analysis and Retrospectives
+
+Once an incident has ended, we want to perform an analysis and conduct a
```
travelar, line 3: > Once an incident has ended, we want to perform an analysis and conduct a retrospective to figure out what we can learn from what happened.

I think it's important to call out a blameless retrospective here. From a psychological perspective, I think the wording brings more participation.

incident-response/overview.md
cblkwell: I have *definitely* seen a communicator role in use during big incidents; the IC should definitely not be handling communication with stakeholders and customers, which can often be a heavy lift. Having a separate person (either a dedicated communicator or the scribe) handle it allows them to focus on actually resolving the incident and not losing context, which should be top priority.

incident-response/analysis.md
cblkwell: Yes, that will definitely be a part of the actual section when I work on it. Right now this is just a placeholder for something that will be done for https://www.pivotaltracker.com/story/show/172926162.

incident-response/overview.md
eeeady: +1 to separate communicator. Did that at my last job and it was such a nice experience to have someone on point for it specifically.

incident-response/analysis.md
cblkwell: Added a some blameless verbiage there.

PR #116, opened 2020-06-12
Reorganizing infrasec section and adding charter

cblkwell:
So, I added the charter to this section and also did a bunch of reorganizing so that things made a little more sense. I think there probably needs to be more work to reorganize this and make it more cohesive, but I had already spent a while doing this and I didn't want to spend *all* day doing. We can take a look at doing some more of that later.


cblkwell:
> I generally approve of this change! I have an optional change... it might be a pain to do though... can we move/rename this whole directory tree to InfraSec? I think that might more accurately match with your changes.
> 
> Thanks Chas!

:blerg: I thought about that but I was hoping to avoid searching for links...I suppose I can be not lazy though. :)

PR #125, opened 2020-07-16
change module env setting to auto

mdawn:
I updated this in `ecs-service-logs` & am updating it here also. With the update to Go version `1.13` we now have an `auto` setting that acknowledges dependencies outside gopath.


mdawn:
Something like `cd && GO111MODULE=on go get` will still have to be used to pick up legacy gopath/vendoring, but afaik that isn't a concern for goreleaser.

chrisgilmerproj:
> Something like `cd && GO111MODULE=on go get` will still have to be used to pick up legacy gopath/vendoring, but afaik that isn't a concern for goreleaser.

You could add this to the docs, but yeah we should be past this need.

PR #126, opened 2020-07-16
'master' to 'main' documentation

chrisgilmerproj:
This documents the transition of `master` to `main` in git while also updating documentation.

[Pivotal story and research](https://www.pivotaltracker.com/story/show/173665753)

### DIFF
--- a/developing/learning/the_effective_engineer.md
+++ b/developing/learning/the_effective_engineer.md
@@ -1,7 +1,7 @@
 # [Learning Resources](./README.md) / "The Effective Engineer"
 
 Edmond Lau's _The Effective Engineer_ is an excellent book that covers many
-aspects of engineering. The master package ($249), available on Edmond's site, is
+aspects of engineering. The main package ($249), available on Edmond's site, is
 worth it for additional material and videos.
 
 ## Recommended by
--- a/developing/learning/the_effective_engineer.md
+++ b/developing/learning/the_effective_engineer.md
@@ -1,7 +1,7 @@
 # [Learning Resources](./README.md) / "The Effective Engineer"
 
 Edmond Lau's _The Effective Engineer_ is an excellent book that covers many
-aspects of engineering. The main package ($249), available on Edmond's site, is
+aspects of engineering. The package ($249), available on Edmond's site, is
 worth it for additional material and videos.
 
 ## Recommended by
--- a/developing/vcs/master-to-main.md
+++ b/developing/vcs/master-to-main.md
@@ -25,7 +25,7 @@ git symbolic-ref refs/remotes/origin/HEAD refs/remotes/origin/main
 git fetch --all --prune
 ```
 
-Test that the `master` branch has been removed by running this command and getting an error:
+Test that the `master` branch has been removed by running this command and getting a shell error (exit code not `0`):
 
 ```sh
 git checkout master
--- a/developing/vcs/master-to-main.md
+++ b/developing/vcs/master-to-main.md
@@ -5,6 +5,7 @@ These instructions will help moving repos from using the default branch name `ma
 
 <!-- toc -->
 
+* [Prereqs](#prereqs)
 * [Instructions to change default branch name](#instructions-to-change-default-branch-name)
 * [Future proofing](#future-proofing)
 * [Github](#github)
@@ -16,6 +17,12 @@ These instructions will help moving repos from using the default branch name `ma
 Regenerate with "pre-commit run -a markdown-toc"
 
 <!-- tocstop -->
+
+## Prereqs
+
+First, ensure that the version of `git` that you have installed is at least v2.28.0. This is important to ensure
+you have access to the [init.DefaultBranch](https://github.blog/2020-07-27-highlights-from-git-2-28/#introducing-init-defaultbranch)
+setting in the git config file.
 
 ## Instructions to change default branch name
 
@@ -47,7 +54,21 @@ git checkout master
 
 ## Future proofing
 
-Until the `git` tool is modified you'll have to be careful to not creat new repos with `master` as the default
+If you're using `git` v2.28.0 or later you should be able to configure git globally to ensure `main` is now
+your default branch permanently.
+
+```sh
+git config --global init.defaultBranch main
+```
+
+Now in your `~/.gitconfig` file you'll see this addition:
+
+```ini
+[init]
+    defaultBranch = main
+```
+
+For older versions of the `git` tool you'll have to be careful to not creat new repos with `master` as the default
 branch. First make a new git alias:
 
 ```sh
--- a/developing/vcs/master-to-main.md
+++ b/developing/vcs/master-to-main.md
@@ -3,6 +3,20 @@
 These instructions will help moving repos from using the default branch name `master` and modify it to be
 `main`.
 
+<!-- toc -->
+
+* [Instructions to change default branch name](#instructions-to-change-default-branch-name)
+* [Future proofing](#future-proofing)
+* [Github](#github)
+* [CircleCI Changes](#circleci-changes)
+* [Terraform Registry](#terraform-registry)
+* [IDE Changes](#ide-changes)
+* [References](#references)
+
+Regenerate with "pre-commit run -a markdown-toc"
+
+<!-- tocstop -->
+
 ## Instructions to change default branch name
 
 To update the branch name use the npm tool [lgarron/main-branch](https://github.com/lgarron/main-branch):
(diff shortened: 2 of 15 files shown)


developing/learning/the_effective_engineer.md
```diff
@@ -1,7 +1,7 @@
 # [Learning Resources](./README.md) / "The Effective Engineer"
 
 Edmond Lau's _The Effective Engineer_ is an excellent book that covers many
-aspects of engineering. The master package ($249), available on Edmond's site, is
+aspects of engineering. The main package ($249), available on Edmond's site, is
```
kahlouie, line 4: I think this one might actually be master 😬

developing/vcs/master-to-main.md
```diff
@@ -0,0 +1,89 @@
...
+git pull origin main
+git checkout main
+git branch --delete master
+git branch --set-upstream-to=origin/main main
+git symbolic-ref HEAD refs/heads/main
+git symbolic-ref refs/remotes/origin/HEAD refs/remotes/origin/main
+git fetch --all --prune
+```
+
+Test that the `master` branch has been removed by running this command and getting an error:
```
kahlouie, line 28: Given that this is a test, what should the output be that I'm expecting.

developing/vcs/master-to-main.md
```diff
@@ -0,0 +1,89 @@
...
+
+Until the `git` tool is modified you'll have to be careful to not creat new repos with `master` as the default
+branch. First make a new git alias:
+
+```sh
+git config --global alias.main '!git init && git checkout -b main'
+git main
+```
+
+You can add an alias to your `bash` or other shell to help:
```
kahlouie, line 44: can you add something about sourcing that work or reloading the shell tab to take effect? I know that's bitten some juniors and under caffeinated folks (like me) before.

developing/vcs/master-to-main.md
```diff
@@ -0,0 +1,89 @@
...
+```
+
+Which gives you the ability to run either `git main` or `git_init` to start a new repo with `main` as the default
+branch.
+
+## Github
+
+The branch protection rules in Github will need to be modified for each repo that you touch.
+
+## CircleCI Changes
```
kahlouie, line 63: Can you look into hosted circleci instances as well? CMS for example is using a private hosted one.

developing/vcs/master-to-main.md
```diff
@@ -0,0 +1,89 @@
...
+## Terraform Registry
+
+Terraform registry uses git tags to pull from github. As long as `main` is a reflection of `master` then no
+changes are needed for the terraform registry.
+
+## IDE Changes
+
+### GoLand
+
+Appears that no changes are required. Consider adding `main` to the set of protected branches.
```
kahlouie, line 83: guidance on that? Maybe also for atom and vscode as I know a few engineers using those as well.

infrasec/aws/aws-organizations.md
```diff
@@ -22,7 +22,7 @@ description of the patterns we've adopted.
 
 * Pick a standard prefix you can use for all the accounts in your
   organization; for our example we'll use `spacecats`.
-* The first account you should create is the "master" account for the
+* The first account you should create is the "main" account for the
```
kahlouie, line 25: While I agree this shouldn't be "master", should this be something else like "primary" as it's referring to a totally different concept from branches?

infrasec/aws/aws-organizations.md
```diff
@@ -108,7 +108,7 @@ experiences on various projects.
   to be `spacecats-infra+aws-org-id@truss.works`, etc. This email will be
   where support responses and other notifications about your AWS accounts
   will be sent.
-* As stated above, *severely limit access to the master account.* This
+* As stated above, *severely limit access to the main account.* This
```
kahlouie, line 111: here too, could be "primary"

developing/learning/the_effective_engineer.md
chrisgilmerproj: Probably I could just remove the word and it would still be clear. Wdyt?

developing/vcs/master-to-main.md
chrisgilmerproj: `getting an error` is the test that validates it.

developing/vcs/master-to-main.md
chrisgilmerproj: I'm thinking that's outside of my scope. Happy to help with it but I'm going to stick to Truss only, non-project changes first. And we can follow up with the CMS folks when we get done with the first part. Same for other projects.

developing/vcs/master-to-main.md
chrisgilmerproj: I don't actually use any editors except `vim`. I reached out to folks who know these editors and assume they can help each other or update these docs. I really don't want to be exhaustive, or alternatively I can leave out any mention of editors.

infrasec/aws/aws-organizations.md
chrisgilmerproj: Honestly any word would work here. `main` is just as good as `primary`. From an infra perspective I don't think I actually use one or the other of those two examples.

developing/learning/the_effective_engineer.md
kahlouie: remove works too!

developing/vcs/master-to-main.md
kahlouie: I think I'd want to know what error I'm expecting or how I know things have gone off the rails :P

developing/vcs/master-to-main.md
kahlouie: I think this is part of the derisking process, but I understand that it should be delayed until after. @jsclarridge maybe you know?

developing/vcs/master-to-main.md
kahlouie: Fair. Removing seems like a better idea (or saying any IDE that handles github integrations should be added to protected, if possible.

developing/vcs/master-to-main.md
chrisgilmerproj: It's just a shell error. I guess I could write `running this command and getting a shell error` if that makes it more clear.

infrasec/aws/aws-organizations.md
brainsik: Kara is bringing up a good point here. This terminology comes directly from AWS and changing it to "main" is  nonsensical since it doesn't correlate with anything else about AWS Organizations nor is it in our control (compared to the default git branch) to change.

That said, what I think what we could do here is use something like "root", "organization root", or "org root". That concept _is_ part of AWS Organizations, so would line up with existing docs and our usage. It might make more sense to break that change out to a different PR so we could get some eyes on that piece from Chas and others who've been dealing a lot with AWS Organizations.

brainsik:
I dropped a comment about how the changes to the AWS Organizations needs a different tactic.

infrasec/aws/aws-organizations.md
chrisgilmerproj: I'll go with `org-root`

developing/vcs/master-to-main.md
```diff
@@ -0,0 +1,125 @@
...
+<!-- toc -->
+
+* [Prereqs](#prereqs)
+* [Instructions to change default branch name](#instructions-to-change-default-branch-name)
+* [Future proofing](#future-proofing)
+* [Github](#github)
+* [CircleCI Changes](#circleci-changes)
+* [Terraform Registry](#terraform-registry)
+* [IDE Changes](#ide-changes)
+* [References](#references)
```
ahobson, line 15: Should we mention that danger `10.3.0` is necessary for `danger local` to work?

chrisgilmerproj: Sure! Can you also send me the PR for that change?

ahobson: This is the [danger issue I created](https://github.com/danger/danger-js/issues/1057) which has a link to the PR that resolves it.

chrisgilmerproj:
It appears that the name chosen is going to be `main`. So I'd like to get this in as a foundation for the work and then we can iterate in future PRs.

developing/learning/the_effective_engineer.md
```diff
@@ -1,7 +1,7 @@
 # [Learning Resources](./README.md) / "The Effective Engineer"
 
 Edmond Lau's _The Effective Engineer_ is an excellent book that covers many
-aspects of engineering. The master package ($249), available on Edmond's site, is
+aspects of engineering. The package ($249), available on Edmond's site, is
```
eeeady, line 4: Lol I'm not sure you needed to scrub it here but okay.

PR #127, opened 2020-07-24
Add Pulumi to the «truss stack»

cobraz:
Hello 👋

Love your playbook! Being a developer and CEO at Bjerk, an Oslo-based development agency, I love to read how other agencies do these kinds of things. Much appreciated!

As a token of gratitude, I'd like to tip you to Pulumi. We've moved most of our Terraform code over to Pulumi, and even made a few Open Source packages for it. Maybe a fit for the «truss stack»?

You probably already know if it, if not – check it out :)

### DIFF
--- a/infrasec/README.md
+++ b/infrasec/README.md
@@ -21,6 +21,7 @@ response](../incident-response/README.md), and a number of other fields.
   - [Orbs](./circleci/orbs.md)
 - Packer
 - [Terraform](./tf/README.md)
+- Pulumi
 
 ## Immutable Infrastructure
 


dynamike:
We've had some conversations about Pulumi internally. The main takeaway we have right now is Pulumi appears to fall into a negative pattern we've seen with Troposphere, where the ability to use a real programming language allowed folks to get to fancy and that ended up hurting the debugging and visibility into how the infrastructure is being managed. 

With that being said, we still need to experiment with it and will update the playbook once we have.

PR #128, opened 2020-07-29
vpc best practices first pass

eeeady:
Adds some basics on vpc best practices and includes some guidance on vpc endpoints... i'm sure i missed something... lol

### DIFF
--- a/infrasec/aws/vpcs.md
+++ b/infrasec/aws/vpcs.md
@@ -1,0 +1,81 @@
+# [AWS](./README.md) / VPC Configuration
+
+## VPC basics
+
+A VPC is a "Virtual Private Cloud".
+This is a virtual network defined in AWS.
+Each VPC is defined in a specified [region](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html#concepts-available-regions).
+
+The VPC needs a defined CIDR block.
+These are private network blocks and need to have enough ip addresses for whatever you plan to deploy.
+When in doubt, use a `/16`.
+
+Note that each region has multiple [Availability Zones](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html#concepts-availability-zones) and the ip space will be subdivided into subnets across these availabilty zones.
+
+The following are objects associated with VPCs that can be configured.
+
+### Subnets
+
+A subnet is a logical subdivision of an IP network.
+Generally, you will create a pair of public and private subnets in each availability zone for the region.
+
+Creating additional subnets will create additional complexity in route tables.
+If your project requires you to isolate resources from one another, use Security groups.
+
+### Route tables
+
+Route tables define the network routing in the VPC.
+You can route specific ip address spaces to separate subnets or to gateways.
+
+### Network ACLs
+
+Network ACLs are like firewall rules applied to specific subnets in a VPC.
+You can define inbound and outbound network traffic across ports for specific IP addresses.
+We prefer to use Security Groups over Network ACLs as they can be applied to specific resources in a VPC.
+
+### Security Groups
+
+In a traditional corporate network these would be equivalent to firewall rules.
+Security Groups allow you to define inbound and outbound network traffic across ports that than be specific to IP addresses OR other security groups.
+Because you can define security group access using security group names, we find it easier to configure intended access to resources than Network ACLs.
+
+### Gateways
+
+For all public subnets in a VPC, use a singlle internet gateway for default routes.
+This means they can route to the public internet.
+
+For each private subnet, create a NAT gateway with an EIP.
+You do not need a NAT instance.
+
+### VPC endpoints
+
+VPC endpoints are AWS provided and defined endpoints for AWS services.
+You may want to use this to close up your network so you have no public traffic.
+
+There are two types of VPC endpoints, Gateway endpoints and Interface endpoints.
+
+Interface endpoints are a deployed elastic network interface with a private IP address connected to your VPC.
+Note: In our internal test configuration we weren't sure it was worth the cost to just spin up one of everything.
+
+Gateway endpoints are existing endpoints that require configuration in the route tables.
+Available Gateway endpoints are: AWS S3 and AWS DynamoDB endpoints in your VPC and you should just deploy those.
+
+The full list of supported endpoints is [here](https://docs.aws.amazon.com/vpc/latest/userguide/vpc-endpoints.html).
+
+### Flow logs
+
+VPC Flow logs allow you to monitor all network traffic in your VPC.
+Unless you have a plan to ingest, manage, view, and monitor network flow logs we do not recommend you turn these on.
+
+## Default VPC configuration
+
+Every AWS account comes with a default VPC.
+In general, you will not want to use that.
+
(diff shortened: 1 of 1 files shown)


infrasec/aws/vpcs.md
```diff
@@ -0,0 +1,81 @@
...
+
+VPC Flow logs allow you to monitor all network traffic in your VPC.
+Unless you have a plan to ingest, manage, view, and monitor network flow logs we do not recommend you turn these on.
+
+## Default VPC configuration
+
+Every AWS account comes with a default VPC.
+In general, you will not want to use that.
+
+Use our [VPC Terraform Module](https://github.com/trussworks/terraform-aws-default-vpc) to configure the default VPC to not allow public network access.
```
chrisgilmerproj, line 75: 🎉

infrasec/aws/vpcs.md
```diff
@@ -0,0 +1,81 @@
...
+When in doubt, use a `/16`.
+
+Note that each region has multiple [Availability Zones](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html#concepts-availability-zones) and the ip space will be subdivided into subnets across these availabilty zones.
+
+The following are objects associated with VPCs that can be configured.
+
+### Subnets
+
+A subnet is a logical subdivision of an IP network.
+Generally, you will create a pair of public and private subnets in each availability zone for the region.
```
chrisgilmerproj, line 20: I suggest we also give some guidance on using either `/22` or `/24` for subnets as a default.

infrasec/aws/vpcs.md
dynamike: I ran into issues with this module on move.mil. It effectively disabled the default VPC, but when I wanted undo that change by removing the module, the change to the default VPC did not get undone and I was forced to undo the changes manually which required using AWS CloudTrail and AWS Config to find and undo the changed made by the module.

infrasec/aws/vpcs.md
```diff
@@ -0,0 +1,82 @@
...
+You do not need a NAT instance.
+
+### VPC endpoints
+
+VPC endpoints are AWS provided and defined endpoints for AWS services.
+You may want to use this to close up your network so you have no public traffic.
+
+There are two types of VPC endpoints, Gateway endpoints and Interface endpoints.
+
+Interface endpoints are a deployed elastic network interface with a private IP address connected to your VPC.
```
dynamike, line 58: I think it would be good to give terraform example fo each of the interface type. It's not 100% straight forward how to do this.

infrasec/aws/vpcs.md
```diff
@@ -0,0 +1,82 @@
...
+In a traditional corporate network these would be equivalent to firewall rules.
+Security Groups allow you to define inbound and outbound network traffic across ports that than be specific to IP addresses OR other security groups.
+Because you can define security group access using security group names, we find it easier to configure intended access to resources than Network ACLs.
+
+### Gateways
+
+For all public subnets in a VPC, use a singlle internet gateway for default routes.
+This means they can route to the public internet.
+
+For each private subnet, create a NAT gateway with an EIP.
```
dynamike, line 48: Should the EIPs be deployed with the NAT Gateway or separate? In some cases, the static IPs are very special and used for firewall in dependent services.

infrasec/aws/vpcs.md
```diff
@@ -0,0 +1,82 @@
...
+
+Gateway endpoints are existing endpoints that require configuration in the route tables.
+Available Gateway endpoints are: AWS S3 and AWS DynamoDB endpoints in your VPC and you should just deploy those.
+
+The full list of supported endpoints is [here](https://docs.aws.amazon.com/vpc/latest/userguide/vpc-endpoints.html).
+
+### Flow logs
+
+VPC Flow logs allow you to monitor all network traffic in your VPC.
+Unless you have a plan to ingest, manage, view, and monitor network flow logs we do not recommend you turn these on.
```
dynamike, line 69: So here's a question, that I don't 100% know the answer. AWS GuardDuty relies on VPC flow logs to be able to detect things like SSH brute force attempts. What I don't know if we need to turn flow logs on for that to work or if AWS somehow behind the scenes is able to use flow logs even if we leave them off.

infrasec/aws/vpcs.md
```diff
@@ -0,0 +1,82 @@
...
+The following are objects associated with VPCs that can be configured.
+
+### Subnets
+
+A subnet is a logical subdivision of an IP network.
+Generally, you will create a pair of public and private subnets in each availability zone for the region.
+A good default CIDR range for each subnet is either a `/22` or `/24`.
+
+Creating additional subnets will create additional complexity in route tables.
+If your project requires you to isolate resources from one another, use Security groups.
```
ahobson, line 24: At OneMedical, they initially reused subnets in various places and Jeremy suggested they would regret it.  By the time I was there, they were regretting it.  Perhaps call out reuse can be a problem?

eeeady: Good point. Going to add that.

infrasec/aws/vpcs.md
eeeady: Uh... the list has gotten kinda long... I'll see what I can do though.

infrasec/aws/vpcs.md
eeeady: Good question. I was going based on our internal doc... I'll give it a think.

Do you have a gut check thought on which makes more sense?

infrasec/aws/vpcs.md
eeeady: /me lies facedown on the floor

Maybe for now I'll throw in this note so folks know we don't know that yet... and maybe put a task in our backlog to research that.

eeeady:
@dynamike I wanna understand why y'all needed to undo the module. I thought we advise against using the default VPC and so that module just disables it. I wanna at least document why you don't want to use it in certain situations here and in that module.

dynamike:
> @dynamike I wanna understand why y'all needed to undo the module. I thought we advise against using the default VPC and so that module just disables it. I wanna at least document why you don't want to use it in certain situations here and in that module.

Yeah, we do want to disable the default VPC generally speaking. In this case the default VPC was being used in by existing live environment that I wasn't aware of so I needed backed out the changes and couldn't. 9 times out of 10 it's probably fine, but there's a risk if we expect terraform to undo a change and it can't. We can pop this in a Zoom, if what I've described isn't helpful.

eeeady:
That makes sense. Thank you @dynamike.

infrasec/aws/vpcs.md
eeeady: https://www.pivotaltracker.com/story/show/174102611 added this to look into it

infrasec/aws/vpcs.md
eeeady: It should only be for outbound... I believe the NAT is provisioned with a public IP so... we'll just go with it shouldbe provisioned with it unless you need otherwise.

eeeady:
@dynamike are my warnings/minor edits sufficient for this pass on documenting this?

eeeady:
@dynamike We renamed the offending repo and I updated the documentation. Is this acceptable?

PR #129, opened 2020-08-19
moar teardown hints

mdawn:
Corresponds to https://github.com/trussworks/legendary-waddle/pull/312

May add some more clarification (or convert to a more succinct statement in the paragraph above it), so opening this PR as a draft.

This has tripped us up a few times so I thought I'd add it. HOWEVER if this goes better in `legendary-waddle` because this bit would be très  obvious to more experienced infra folk in our wider audience, lmk. 😛


ralren:
Can we also add that, in order to add billing, one must try to leave the organization to get that prompt? @cblkwell

mdawn:
Ok. Now it's similar to the step above, but for OUs instead of resources.

PR #136, opened 2020-09-19
Major InfraSec cleanup

brainsik:
* `pre-commit autoupdate` + corresponding changes to TOC generation
* Renamed `tf` directory to `terraform`
* Follow the existing pattern and consolidate images into `images` directories
* Broke out "Intro to CI/CD" into it's own article
* Moved "CircleCI Orbs" article to CI/CD directory
* Refined AWS Naming article
* Added TOCs to big articles
* Dropped leading `./` from relative links
* Made opening sections more consistent in how they start
* Made casing for section headers more consistent [unfinished]

PR #137, opened 2020-09-22
Add _Good Infrastructure_ philosophy article

brainsik:
~This PR includes the changes in #136. Once that lands, the only commit left will be the final one which adds the philosophy article.~ Landed.

This pulls in the philosophy section from the trussworks/good-infra  repo.

### DIFF
--- a/infrasec/good-infra.md
+++ b/infrasec/good-infra.md
@@ -82,6 +82,8 @@ Examples:
 
 Codify entire stacks into a single callable unit (calling into other units). Keep the number of parameterizations (inputs) small so the infrastructure doesn't vary wildly and the unit's actions remain clear. This makes it easy to spin up the same infrastructure in multiple environments and reduces the proliferation of code unique to each environment.
 
+Care should be given to not hardcode environmental assumptions into this wrapper. Follow the same notions for building a [Twelve-Factor App](#the-twelve-factor-app). The callable stack is our version of a deployable artifact.
+
 ## Delivery
 
 ### Infrastructure as code
@@ -163,7 +165,7 @@ An effect of adopting an immutable architecture is being guided toward more mini
 
 ## The Twelve-Factor App
 
-Build infrastructure to support [⎋the Twelve-Factor methodology](https://12factor.net/). We should be architecting our delivery and deployment pipelines with this view and pushing back against changes the undermine this methodology.
+Build infrastructure in the spirit of and to support [⎋the Twelve-Factor methodology](https://12factor.net/). We should be architecting our delivery and deployment pipelines with this view and pushing back against changes which undermine this methodology.
 
 [^1]: There's no single strategy. Managing slow tests is complicated topic. You need to weigh the risks of getting slow feedback for each component to arrive at a place that feels comfortable.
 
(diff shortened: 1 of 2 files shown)


brainsik:
Pulled in the changes from #136 landing.

ralren (approved):

infrasec/good-infra.md
```diff
@@ -0,0 +1,170 @@
...
+
+Examples:
+
+* Parameter Store instead of Hashicorp Vault
+* Fargate instead of EC2
+* SaaS versus on-premise
+
+### Codify entire stacks
+
+Codify entire stacks into a single callable unit (calling into other units). Keep the number of parameterizations (inputs) small so the infrastructure doesn't vary wildly and the unit's actions remain clear. This makes it easy to spin up the same infrastructure in multiple environments and reduces the proliferation of code unique to each environment.
```
cblkwell, line 83: I wonder if it's worth noting that it's often good to provide the *ability* to input tweaked parameters, but have sensible defaults set. That's one thing I ran into with MilMove, when we had to move everything into GovCloud, is that there were a lot of things that had been hardcoded that would have been a lot easier if there was a way to just tweak the module from the calling point. Obviously there are tradeoffs here, but..

cblkwell: Also, there is a discussion to be had here about how large a module gets before it becomes unwieldy. An eye towards being able to modify the code in the future is something that needs to be taken into account.

cblkwell: I wonder if this is the place to say something about not encoding the notion of environments or other artificial constructs in the code...

cblkwell (approved):
Looks good to me, really, just a couple of comments in one section.

infrasec/good-infra.md
eeeady: Generally agree with what Chas is saying. I think it's a good spot to put some best practice notes and philosophy behind them here. Specifically because we know they come up.

infrasec/good-infra.md
```diff
@@ -0,0 +1,170 @@
...
+
+In more sophisticated systems, where possible (e.g., web services), the code deploy and the feature release should be separated via feature flags.
+
+## Immutability
+
+> The entropy of the universe tends to a maximum.
+>
+> -- Rudolf Clausius, 1865
+
+Our systems tend to disorder over time. Worse, they do this in non-uniform ways. There are various mechanisms for helping long-lived systems stay in their expected states (e.g., orchestration), but it is somewhat of a fools errand since there's always parts you aren't tracking and stateful bits which are necessarily different. Include the will of an adversary to purposely change the system in a non-detectable way and you may begin questioning your career choices (or worse, stop caring).
```
eeeady, line 148: preach

eeeady (approved):
I generally like these changes. I do think Chas' suggestions of putting some thoughts around how big a tf module should get or setting good default values in vars are important to document... perhaps those are good infra practices just around tf where this is more like general philosophy to take to whatever tech we use?

infrasec/good-infra.md
brainsik: I'll take another stab at this. The idea here is to have a lightweight wrapper for a whole stack. I purposely don't use the word "module" because maybe this is a script or tool. The point is to provide an easy way to deploy the same stack to any environment with minimal difference. I've been wondering if this is our (InfraSec) version of a deployable "artifact". That's the kind of thinking I'm doing around it.

brainsik: Alright. I added a little bit about treating these as Twelve-Factor apps. I'm happy to have a bigger discussion and revise accordingly! I'd like to hear what happened on MilMove, that might help me understand what else needs to be said.

PR #138, opened 2020-09-23
Add a how-to on how to fix the circleci integration

ralren:
We should document on how to fix the circleci integrations.


eeeady (approved):
THANK YOUUU

PR #139, opened 2020-09-29
Add section on Docker IDs and Digests

jsclarridge:
Some recent conversations about the differences between Docker Image IDs and sha256 digests, especially the differences between locally built images and images pushed to a registry, revealed that it would be helpful to document this more thoroughly. I'll admit I had only a loose understanding of the differences between the different IDs and digests, and I wanted to write this up in part to solidify my grasp of the topic. For now I've put this in the Docker section of the playbook, but I'm open to suggestions whether some of the content should be moved to or cross-referenced from the CI/CD section.

### DIFF
--- a/developing/docker/README.md
+++ b/developing/docker/README.md
@@ -11,6 +11,10 @@
 * [Volume Mount Performance](#volume-mount-performance)
 * [Inspecting the Docker Virtual Machine](#inspecting-the-docker-virtual-machine)
 * [Inspecting Container Resources](#inspecting-container-resources)
+* [Image IDs, Tags, and Digests](#image-ids-tags-and-digests)
+  * [Problem Statement and Summary](#problem-statement-and-summary)
+  * [Deep Dive](#deep-dive)
+  * [Locally Built Images vs. Images Pushed to a Registry](#locally-built-images-vs-images-pushed-to-a-registry)
 * [Cleaning Up](#cleaning-up)
 * [Docker Configuration](#docker-configuration)
   * [Keybindings](#keybindings)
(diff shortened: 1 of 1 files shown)


eeeady (commented):
Thank you for writing this up Josh! Sorry it took me a bit to get through and think about.

Generally I approve BUT I would really appreciate it if you have the answer to your problem statement explicitly stated.

This is a lot of good/useful information but I'm not sure what I should be tagging images as so I can deploy "the right one".

developing/docker/README.md
```diff
@@ -94,6 +98,162 @@ To get a `top` like report of what your containers are doing
...
+An [ECS task
+definition](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#container_definition_image)
+can pull an image by tag or digest, so which should you use? Image tags are
+potentially mutable and are not tied to the image contents. For instance, a new
+image could be created with different contents, but could reuse the same tag as
+an earlier, different image. If your ECS task definition pulls the image based
+on tag, you could be pulling in a different image than expected. An image's
+digest, on the other hand, is a hash based on the image contents. Pulling the
+image based on digest provides greater security and protection against
+unintended changes.
```
ahobson, line 119: You can configure ECR with [Immutability](https://docs.aws.amazon.com/AmazonECR/latest/userguide/image-tag-mutability.html) so you cannot overwrite a tag.  IMO, we should always enable that.  That does make it pretty impossible do things like having an up to date `dev`, `production` or `latest` tag.  Again, IMO, you shouldn't be using those anyway.

developing/docker/README.md
```diff
@@ -94,6 +98,162 @@ To get a `top` like report of what your containers are doing
...
+
+When building a Docker image locally, the SHA256 digest to use in the ECS task
+definition won't be known until the image has been pushed to the Docker
+registry. The reason for this is explained in more detail in the sections
+below.
+
+However, once the image has been pushed to the registry, the SHA256 digest to
+use in the ECS task definition can be found with the following command:
+
+    docker image inspect some_image_name -f '{{.RepoDigests}}'
```
ahobson, line 129: Note that `RepoDigests` is an array, so if you have perchance pushed to more than one repository, you will need to make sure you get the digest for the right repository.

developing/docker/README.md
```diff
@@ -94,6 +98,162 @@ To get a `top` like report of what your containers are doing
...
+    Layer: identified by a digest, which is a computed SHA256 of the layer's
+    content. If the content changes, the digest will change.
+
+    Image configuration object (JSON): includes, among other things, an ordered
+    list of all the layer digests that combine to form an image.
+
+Now, let's go through some of the fields in the `docker image inspect` output:
+
+    Id: this is the Image ID, a digest or computed SHA256 hash of the image's JSON
+    configuration object. Note: this is different from the RepoDigest.
```
ahobson, line 215: The JSON configuration object contains information about the local machine and thus that is why the `Id` is not globally unique.

ahobson:
> Thank you for writing this up Josh! Sorry it took me a bit to get through and think about.
> 
> Generally I approve BUT I would really appreciate it if you have the answer to your problem statement explicitly stated.
> 
> This is a lot of good/useful information but I'm not sure what I should be tagging images as so I can deploy "the right one".

I think if you need/want "end to end" verification, then after pushing the image you save the `RepoDigest` and then use that during the deploy phase.  I believe that's what Chas ended up doing on MilMove.

PR #140, opened 2020-10-06
Initial draft of configuration docs.

jim:
This content in this PR is a summary of [this Slack thread](https://trussworks.slack.com/archives/CDJ9JAD26/p1601998134067100).

### DIFF
--- a/developing/configuration/README.md
+++ b/developing/configuration/README.md
@@ -3,11 +3,15 @@
 Projects should adhere to the following guidelines for managing configuration:
 
 1. Configuration values should be provided to programs via environment variables.
-2. Programs should validate configuration at boot time.
+2. Programs should validate configuration immediately when they are run.
 3. If config is incomplete or invalid, programs should:
    1. Exit with a non-zero status
    2. Print a helpful error message to assist humans in resolving the issue
-4. All environment values should be cataloged in a project's `.envrc` file, although values can be left blank. The purpose of this file is to document all possible configuration points.
+4. All environment values should be cataloged in a single location.
+
+   - This should generally be a project's `.envrc` file (values can be left blank as needed), although ecosystems that have a strong convention for this configuration living elsewhere should follow that convention.
+   - The purpose of maintaining this catalog of environment variables is to document all possible configuration points.
+
 5. Projects should define default configuration for local development in an `.envrc.local` file. This file should be loaded from the project's main `.envrc` file if it exists. [Here is an example](https://github.com/CMSgov/easi-app/blob/master/.envrc#L77-L80).
 
 Following the above guidelines should yield projects with a setup that allows developers to get a new project clone running with minimal manual configuration.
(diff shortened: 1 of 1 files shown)


developing/configuration/README.md
```diff
@@ -0,0 +1,13 @@
+# Configuration
+
+Projects should adhere to the following guidelines for managing configuration:
+
+1. Configuration values should be provided to programs via environment variables.
+2. Programs should validate configuration at boot time.
```
pkarman, line 6: is it maybe worth clarifying `program boot time` to distinguish from "OS boot time"?

pkarman (commented):
one editorial suggestion.

Is it maybe worth scoping the `.envrc` convention to include a caveat "unless the language/framework already has a well-established convention for managing environment variables". This is the first time I've seen `.envrc` and while it's cool for Golang I just don't know if other server-side apps (e.g. Rails) might have other well-known conventions.

macrael:
@pkarman FWIW .envrc is not at all a Golang convention, it's a truss-infra convention. I think in part because it is language agnostic, so you can use it across both services and tools.

developing/configuration/README.md
cblkwell: Isn't this "run time" (ie, when the program is run)? Or am I misunderstanding a nuance?

jim: How about "programs should validate configuration when they are run"? I was trying to communicate that it should occur immediately when the program starts and not at some point later on.

developing/configuration/README.md
```diff
@@ -0,0 +1,17 @@
...
+Projects should adhere to the following guidelines for managing configuration:
+
+1. Configuration values should be provided to programs via environment variables.
+2. Programs should validate configuration immediately when they are run.
+3. If config is incomplete or invalid, programs should:
+   1. Exit with a non-zero status
+   2. Print a helpful error message to assist humans in resolving the issue
+4. All environment values should be cataloged in a single location.
+
+   - This should generally be a project's `.envrc` file (values can be left blank as needed), although ecosystems that have a strong convention for this configuration living elsewhere should follow that convention.
```
rpdelaney, line 12:
```suggestion
   - This should generally be a project's `[.envrc](../direnv/README.md)` file (values can be left blank as needed), although ecosystems that have a strong convention for this configuration living elsewhere should follow that convention.
```

rpdelaney (approved):
FYI I think we are no longer requiring peer review to make changes to the eng playbook.

developing/configuration/README.md
rpdelaney: I _think_ that will do what I think it will do... :)

macrael:
I love this, it would be nice to see some examples of the level of description you like for environment variables. On my current project I stole having a separate Configuration.md file that has a couple sentences about each variable in it. I really like your suggestion that that documentation live in .envrc, where all the default vaules for development are going to be laid out, and I think an example would help set the tone for just how much explaination each env var deserves.

PR #142, opened 2020-10-15
update yubikey documentation

carterjones:
This PR is largely composed of three main changes: 1) formatting, 2) content modifications, and 2) content additions.

The formatting includes things like indentation fixes, changing unordered lists to numbered lists, fixing capitalizations, removing unneeded portions from URLs, and resizing images.

The content modifications include indicating some steps as optional, recommending GPG Suite instead of the standard gnupg package, and suggesting a new way to configure gpg-agent on systems for use with YubiKeys.

The content additions include new sections for verifying configuration settings, information on how to delete secret key material, a section for how to export public key, a section explaining how user IDs are stored in the primary key, and a section about subkey stub files.


carterjones:
I'm looking for feedback about the things proposed with this PR. If it seems like this is worth incorporating, I can turn this PR into a regular PR rather than a draft PR.

PR #149, opened 2021-01-15
Include notes on adding AWS support in GovCloud

dynamike:
It's a pain, but I fought the battle to hopefully make it easier for those who want to enable GovCloud AWS Support in the future. Take my money!


cblkwell (approved):

esacteksab (approved):

mdawn (approved):
>  Yes you read that right

💀

PR #151, opened 2021-01-21
Publish terraform modules as trussworks-infra

rpdelaney:
I just published a module this way and it wasn't hard. I figured this would reduce the bus factor of managing our modules in the registry. Tagging for review since I'm not certain the `trussworks-infra` github account is the best for that or not.


cblkwell (commented):
This seems like a reasonable way to go to me, though we probably want more folks to weigh in.

rpdelaney:
I merge based on that and the thumbs up I got in the slack thread from Eeady and Barry

PR #161, opened 2021-02-08
SQL observability

nelztruss:


developing/languages/README.md
```diff
@@ -58,3 +59,11 @@ Important JS patterns and features to understand:
...
 * React and Typescript
   * [Truss resources for React, Typescript, and front end tooling](https://github.com/trussworks/Engineering-Playbook/tree/master/web/frontend)
+
+### SQL
+
+Did you know that SQL dialects have provisions for leaving comments within their queries? The following observability hint comes directly from [Database Reliability Engineering](https://www.oreilly.com/library/view/database-reliability-engineering/9781491925935/?sortby=bestSellers):
+
+> When doing SQL tuning, a big challenge is mapping SQL running in the database to the specific place in the codebase from which it is being called. In many database engines, you can add comments for information. These comments will show up in the database query logs. This is a great place to insert the codebase location.
+
+Additionally, if you have a request-scoped identifier (e.g. a `requestID` or `traceID`), you might be able to weave that into a built SQL query to give even finer-grained visibility.
```
ahobson, line 69: This is really great advice! Do you think adding a concrete example of how to do this would be useful?

Also, for Postgres, I've found that using [application_name](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-PARAMKEYWORDS) in the connection string can be really useful if you have multiple types of connections (e.g. the webapp and an async job queue worker)

nelztruss: Yah, I can add some more links and code and query log output examples.

As I haven't played with it, I'm gonna let you add your `application_name` hint as a separate PR. 😉

ahobson (approved):
Awesome

PR #163, opened 2021-02-11
Add note about module version missing in registry

jsclarridge:
I had a need to do this yesterday and my first thought was to look in the Engineering Playbook. I feel like this question has come up several times and it'd be helpful to add a brief note on how to address it.


cblkwell (approved):
lgtm

PR #173, opened 2021-03-24
Document CircleCI Diff Tool

brandonlenz:
Document an approach to adding logic to CircleCI pipelines to determine the changes from a given PR. The example use case is for use within a monorepo where certain file changes should not require build + test of all the code. This would allow pipelines to intelligently select what to do based on what changed.

For additional context, please see:
 - The raised issue: https://github.com/trussworks/appeng/issues/44
 - A test project PR where I was learning CircleCI and testing this job: https://github.com/brandonlenz/circle-ci-test-repo/pull/2

Personally, I get the feeling that this would be useful as a reusable utility in the form of an Orb, but as far as I can tell CircleCI does not support returning values from an Orbs job to then be used as input parameters for other jobs or commands. The diff can be stored as a BASH_ENV variable and passed to different steps, or maybe shared between workspaces as far as I understand, however.

I was also a bit unsure of where to put this information. I'm good to move it if someone is aware of a better location. Creating anything "utilities" or "utils" related is normally a code smell, so I'm not feeling confident, but didn't know what else to call this.

### DIFF
--- a/developing/cicd/README.md
+++ b/developing/cicd/README.md
@@ -10,6 +10,7 @@ according to Truss best practices to deliver software to customers.
 * [Standard Delivery Pipeline](delivery-pipeline.md)
 * [CircleCI Config Patterns](circleci-patterns.md)
 * [CircleCI Orbs](circleci-orbs.md)
+* [CircleCI Utilities](circleci-utilities.md)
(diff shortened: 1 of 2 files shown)


ahobson (approved):
Thank you, this is great

PR #178, opened 2021-04-13
Just adding a bit of color to our editor lists.

macrael:
I was scanning the playbook and saw some stale data


monfresh (approved):
:shipit:

PR #181, opened 2021-06-01
document context as first param

nelztruss:
I had originally added this bit at https://github.com/trussworks/go-sample-api/tree/main/pkg#context... But now adding here, too, for visibility.

### DIFF
--- a/developing/languages/GO.md
+++ b/developing/languages/GO.md
@@ -2,7 +2,7 @@
 
 ## Overview
 
-Go aka Golang is a programming language designed by folks at Google.
+Go (a.k.a. Golang) is a programming language designed by folks at Google.
 
 This guide provides some general resources for working in Go.
 [Web Application in Go](../../web/server/go.md)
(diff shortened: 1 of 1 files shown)


developing/languages/GO.md
```diff
@@ -58,6 +58,14 @@ provides some specifics to web development.
...
   it might mean it needs to be refactored as it's exported function elsewhere.
 
+## Context
+
+### Function First Parameter
+
+As a general rule, `context.Context` should be passed down through the layers of your program, as this is the conventional Go way [to address "cross cutting concerns"](https://blog.golang.org/context-and-structs#TOC_5.), e.g. cancellation, logging, distributed tracing, or other types of instrumentation (which in other languages might be addressed via Thread Locals or similar constructs).
+
+At Google, they found this pattern to be so useful that [they require it](https://blog.golang.org/context#TOC_5.).
+
```
ahobson, line 68: This is great!

Do you think it would be worthwhile digging into when (if ever) that context should be saved to a receiver struct? I'm trying to think of questions or confusion that someone might have.

nelztruss: That's actually called out in the conclusion section of the "context-and-structs" blog post:

> When designing an API with context, remember the advice: pass `context.Context` in as an argument; don't store it in structs.

Do you think I should I pull that out to re-emphasize it here?

ahobson: Thank you for pointing out what I missed. Nah, let's leave it as is. We can always add more clarification in the future if needed

ahobson (approved):

PR #184, opened 2021-06-16
cross-language interop

nelztruss:
Advice to match identifiers to the destination representation.

### DIFF
--- a/developing/languages/README.md
+++ b/developing/languages/README.md
@@ -22,6 +22,31 @@ While Truss may have opinions on particular languages and strategic reasons for 
 
 ## Language Resources
 
+### Cross-Language Patterns
+
+In many projects, it is fairly common to have interoperation between disparate
+languages, e.g. a Python codebase that also builds an OpenAPI spec or Go
+entities that are marshaled into JSON.
+
+As a general guideline, it is recommended to use identifiers that are idiomatic
+to the destination representation.
+
+Some things to consider:
+
+* Swagger/OpenAPI specs are intended to be JSON (or YAML), so whether you're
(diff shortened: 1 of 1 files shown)


developing/languages/README.md
```diff
@@ -22,6 +22,31 @@ While Truss may have opinions on particular languages and strategic reasons for
...
+
+Some things to consider:
+
+* Swagger/OpenAPI specs are intended to be JSON (or YAML), so whether you're
+coming from Python (`snake_case`) or BASIC (`CAPITAL_CASE`) or Go
+(`PascalCase`), prefer to output JSON with idiomatic `camelCase` identifiers.
+* Structured logging usually uses JSON as the destination structure, so when
+creating log attributes/names, prefer JSON idiomatic identifiers in
+`camelCase` where possible.
+* Go's idiomatic "acronyms should be capitalized" is actually lossy when it
```
kimallen, line 42:
```suggestion
* Go's idiomatic "acronyms should be capitalized" is actually lousy when it
```

felipe-lee: I guess I'm not sure if it's meant to be lossy as in you lose some data when taken to this new context because it thinks the whole thing should be split up into one, or lousy like you corrected, because it can't tell what it should be converted to.

jim (approved):
Is there a general recommendation to use camelCase with any API that speaks JSON? That's what I took away from this, but if that is accurate, I think we could make that more explicit or a little TLDR at the top of this section.

developing/languages/README.md
nelztruss: Funny!
Spoke with Kim, gonna keep the original usage with this meaning: https://techterms.com/definition/lossy

PR #187, opened 2021-07-16
Add more info around running nix commands

felipe-lee:
In particular add more info on profiles, working with packages, and installing `nix`.

### DIFF
--- a/developing/nix/README.md
+++ b/developing/nix/README.md
@@ -322,3 +322,11 @@ and break `nix` commands. See creating profiles section for more info.
 
 See the [nix profiles docs](https://nixos.org/manual/nix/stable/#sec-profiles)
 for more info.
+
+## Uninstalling
+
+Uninstalling `nix` takes a few steps:
+
+1. Remove the `nix` entry from `fstab` using `sudo vifs`
+1. Destroy the data volume using `diskutil apfs deleteVolume 'Nix Store'`
+1. Remove the `nix` line from `/etc/synthetic.conf` (w/`sudo`)
--- a/developing/nix/README.md
+++ b/developing/nix/README.md
@@ -77,48 +77,210 @@ docker and a native development environment.
 
 ### Differences with homebrew
 
-Homebrew has more packages than nix, and installs them globally. For
+Homebrew has more packages than `nix`, and installs them globally. For
 certain things, that might work great. However, that doesn't give
 projects reproducible installs.
 
+One other difference is that versions of particular packages may be more
+up to date in one or the other. E.g. as of this writing, homebrew's 
+`watchman` version is `2021.06.07.00` but `nix` is on `4.9.0`, which is 
+from sometime in 2017 or 2018. 
+
 ## Installation
 
+MacOS requires that we add an argument to the installation command. We also want to 
+do the single user installation. So to install, run:
+
+```shell
+sh <(curl -L https://nixos.org/nix/install) --darwin-use-unencrypted-nix-store-volume --no-daemon
+```
+
 See the [macOS installation](https://nixos.org/manual/nix/stable/#sect-macos-installation)
-instructions.
+and [single user installation](https://nixos.org/manual/nix/stable/#sect-single-user-installation) 
+instructions for more details.
 
 ## Getting started
 
-After installing nix, you should have `~/.nix-profile/bin` in your
-PATH. If you're using the fish shell, check out
-[nix-env.fish](https://github.com/lilyball/nix-env.fish). If you are
-using fish, add this to your `~/.config/fish/config.sh`
+After installing `nix`, you should have `~/.nix-profile/bin` in your
+PATH. 
+
+### Extra Setup (Only Fish Shell Users)
+
+If you're using the fish shell, check out
+[nix-env.fish](https://github.com/lilyball/nix-env.fish). 
+
+Add this to your `~/.config/fish/config.sh`
 
     set -g fish_user_paths "/Users/[your_os_username]/.nix-profile/bin" $fish_user_paths
 
-Now you want to create a new profile and install packages in that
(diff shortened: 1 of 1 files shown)


developing/nix/README.md
```diff
@@ -77,48 +77,211 @@ docker and a native development environment.
...
+nix-env -p /nix/var/nix/profiles/milmove -i nix nss-cacert
+```
+
+The `-p` is telling it the path to the profile you want to use for
+this `nix` command. This can be used to work with profiles that aren't
+the active profile. It can also be used with profiles that don't yet
+exist (like here) or with exising profiles.
+
+You may remember the `-i` is used for installing, so we're installing
+`nix` and `nss-cacert` into our new profile.
```
ahobson, line 212: I recommended this at one time, but now I'm not so sure. You can have multiple profiles in your `PATH` and I think that's the better approach: to have the project profile first in `PATH` and the default `~/.nix-profile/bin` later in the `PATH`. That way you can have `nix` and friends in the default profile only.

Also note that you can set `NIX_PROFILE` to point to the current profile path so you don't need to pass `-p`.

developing/nix/README.md
```diff
@@ -77,48 +77,211 @@ docker and a native development environment.
...
+```
+
+E.g.
+
+```shell
+nix-env -S /nix/var/nix/profiles/milmove
+```
+
+Note that it is possible to switch to a profile that doesn't exist yet
+and break `nix` commands. See creating profiles section for more info.
```
ahobson, line 237: IMHO, this shouldn't be necessary. We should be using `NIX_PROFILE` in `.envrc` to switch profiles

developing/nix/README.md
felipe-lee: I'm seeing this a bit differently than you I think. Like I'm seeing this as people being able to have 3 different kinds of profiles:

1. Global profile, which is the default set up by `nix` when it's installed at `/nix/var/nix/profiles/per-user/<username>/profile/`
2. Global project profile, for common things needed across repos for a project. 
    1. E.g. for Milmove, we have multiple repos that each have their own dependencies, but several have some similar base requirements like `aws-vault`, `chamber`, etc.
    2. These can live at 
        `/nix/var/nix/profiles/`
        or
        `/nix/var/nix/profiles/per-user/<username>/`
3. Repo profiles. These would be defined in the repo with the `nix/default.nix` and use env vars set up in `.envrc`

My main reason for distinguishing between 1 and 2 above is that some people may work on other things locally that would need the same tools I define in 2, but different versions. So that would be a use case for having a "global project profile" so that if you are working on a project that uses different versions of tools for their repos, you can just switch global profiles and be set. Then the repo profiles kick in once you get into the repo directory. Maybe I'm over-complicating this though since I'm not sure there are many people that would need to work on different projects. 

Though I do like the idea of not having to install `nix` and `nss-cacert` into all profiles so if we keep the 3 profiles separated like I was thinking, I could update it to include guidance on having your regular project global inserted before the default profile in the `PATH`.

developing/nix/README.md
felipe-lee: This is more if you want to manage profiles at a level outside of a repo. In repos, I think making `.envrc` take care of managing profiles is good.

developing/nix/README.md
ahobson: IMHO, trying to solve the global project profile (use case 2) is a need that we haven't really proven yet. If we had every repo have its own nix config, yes we would have to update multiple repos when a new version of e.g. `aws-vault` came out, but I don't think we have enough experience to know if the additional complexity is worth it.

Maybe with more time we will know more.

Thoughts?

felipe-lee: Yeah, i'm good with keeping it simpler and then re-visit if we come across more concrete needs to change things.

developing/nix/README.md
ahobson: Again, for simplicity's sake, I'm not convinced we need folks worrying about switching repos and breaking things.

developing/nix/README.md
ahobson: How about we move this content to an "advanced users" section or something with a preface about breaking things and invalidating your warranty? 😁

developing/nix/README.md
```diff
@@ -77,48 +77,211 @@ docker and a native development environment.
...
+from sometime in 2017 or 2018.
+
 ## Installation
 
+MacOS requires that we add an argument to the installation command. We also want to
+do the single user installation. So to install, run:
+
+```shell
+sh <(curl -L https://nixos.org/nix/install) --darwin-use-unencrypted-nix-store-volume --no-daemon
+```
```
ahobson, line 96: Thank you so much for adding this clarification

developing/nix/README.md
felipe-lee: That sounds like a good plan. I'll move it instead of deleting it and simplify this part.

developing/nix/README.md
```diff
@@ -162,86 +162,32 @@ nix-env -e chamber
...
+#### Repo Profiles
 
-E.g.
+Ideally you'll have a profile defined at a repo level that will house
+your repo's dependencies. Then you can pre-pend this repo profile to
+your `PATH` when you're working with the repo, while keeping your
+default profile in the `PATH` behind the repo profile to still have
+access to `nix` commands. See
+[Working with Existing Nix Expressions](#working-with-existing-nix-expressions)
+for more details on how to set this up.
```
ahobson, line 188: I like this! How would you feel about documenting the `NIX_PROFILE` environment variable here?

felipe-lee: Ah good point, I'll add that.

ahobson (approved):
THANK YOU! Awesome stuff

PR #189, opened 2021-08-11
Draft: Creating AppEng ADR Process

mikena-truss:
@ntwyman and @ahobson here are some initial thoughts at what structure could look like for bringing in an ADR process to the practice. My assumption is we would do a "dry run" with some clear decisions that we've neglected to standardize in the past for the first quarter and go from there. This would give us the opportunity to experiment.

Feedback needed is mostly around:
* General structure. Is it too lightweight or too structured?
* Time commitments. Will this as actually get done given our time constraints?
* Group structure. Does this breakdown make sense?

Open to all feedback though and can go over in a meeting together.

### DIFF
--- a/practices/appeng/README.md
+++ b/practices/appeng/README.md
@@ -7,7 +7,18 @@ including and mostly dealing with web development.
 This section includes resources
 the practice uses to organize around.
 
-* [Team Charter](team-charter.md) How we set direction as a practice
+## Organizational Resources
+
+Tools we use to organize our people and projects
+
+* [Team Charter](team-charter.md)--what principles drive our engineering culture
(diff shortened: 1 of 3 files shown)


ntwyman:
Generally, I think this looks good.  My one question is about the thinking behind the quarterly review process - is there a particular concern you were trying to address by making this quarterly? My immediate reaction is why batch these up, rather than permit them to be reviewed/accepted at anytime, but I wonder what I am overlooking.

mikena-truss:
Mostly just that we already have some meeting structure in place and likely the leads will be the folks most involved in these decisions. If we'd prefer groups self organize around decisions and the call for async feedback to be enough, I think we could simplify to exactly that. Let's chat next week and settle on which way to go.

ntwyman (approved):

PR #194, opened 2021-09-17
Add nix examples (and follow those examples for the playbook)

ahobson:


.envrc
```diff
@@ -0,0 +1,30 @@
+# Nix config
+
+# if nix is installed, use it
+if [ ! -r .nix-disable  ] && has nix-env; then
+    export NIX_PROFILE=/nix/var/nix/profiles/eng-playbook
+    PATH_add ${NIX_PROFILE}/bin
+    nix_file=nix/default.nix
+
+    unset NIX_SSL_CERT_FILE
```
gidjin, line 9: Since most projects don't need this as noted in the How to, I wonder if we should comment it out instead of leave it in?

ahobson: Oh, yeah! done!

PR #195, opened 2021-09-21
Add mdformat pre-commit hook

rpdelaney:
[mdformat](https://github.com/executablebooks/mdformat) is an opinionated markdown formatter. The [style guide](https://mdformat.readthedocs.io/en/stable/users/style.html) goes over most of what it enforces and why.

Although it has few configuration options, mdformat also supports a [configuration file](https://mdformat.readthedocs.io/en/stable/users/configuration_file.html). I have not added one in this PR, but if we want to enforce word wrap or the other numbering style, we can.

I've left in the markdownlint hook since I'm not certain that mdformat covers everything it looks for. Maybe we don't want both though, in case there are conflicts?


rpdelaney:
@cblkwell 

> Is there a way we can add this WITHOUT it trying to go to enforcement on every single file at once? This is hard to review in any useful way. If we could add it and have it enforce as we update pages that would be a little easier.

I can undo the changes to the markdown files and leave only the pre-commit hook, but I'd expect the CI runs will start to fail because the CircleCI config does this:

```console
pre-commit run --all-files --show-diff-on-failure
```

cblkwell:
I guess this is fine then...I just...I dunno. Don't like this as a general method of operation. :P Also not sure what need this is addressing that the linter was not (but I guess automatically fixing formatting instead of just complaining is nice?)

rpdelaney:
> I guess automatically fixing formatting instead of just complaining is nice?

That's about the extent of it, yeah.

rpdelaney:
Agreed

developing/cicd/delivery-pipeline.md
```diff
@@ -70,16 +68,16 @@ it. This workflow is orchestrated via an automated tool like CircleCI, and
...
+1. CircleCI checks to make sure the deploy was successful. Is our
    environment running the right version of the code?
-4. CircleCI (or another tool fired by CircleCI) performs post-deployment
+1. CircleCI (or another tool fired by CircleCI) performs post-deployment
    testing. Does it pass some functional tests to ensure user workflows
    are functional (eg, can someone log in, pull up a user record, put
    things in a shopping cart, etc)? Are logs filling up with error now
    that the new version is deployed?
-5. If the tests pass, CircleCI goes ahead and deploys to our `production`
+1. If the tests pass, CircleCI goes ahead and deploys to our `production`
```
monfresh, line 80: Why is mdformat flagging these ordered lists? Having them all be `1` is not readable. I've noticed multiple instances of this change. Have the changes that mdformat made been reviewed?

monfresh: It sounds like they're doing it to make diffs easier to read: https://mdformat.readthedocs.io/en/stable/users/style.html#ordered-lists

I don't agree with that reasoning. It's the end result that matters since we're producing material for humans to read. We should optimize for readability of the final product, not of the Git diffs. I wonder what others think. Could we solicit feedback before making big changes like this in the future?

It looks like mdformat has a config to allow consecutive numbering. I would vote to use it.

developing/command-line-tools/README.md
```diff
@@ -88,7 +88,7 @@ You may also be interested in these other pages from the Truss Engineering Playb
 
 ## Document format conversion
 
-- [dos2unix](https://waterlan.home.xs4all.nl/dos2unix.html) Efficient conversion of line terminators between DOS <-> Unix format
+- [dos2unix](https://waterlan.home.xs4all.nl/dos2unix.html) Efficient conversion of line terminators between DOS \<-> Unix format
```
monfresh, line 91: Why is this backslash here?

developing/direnv/README.md
```diff
@@ -8,7 +8,9 @@
 
 Via homebrew:
 
-     brew install direnv
+```
+ brew install direnv
```
monfresh, line 12: There is an extra space here before `brew`.

developing/docker/README.md
```diff
@@ -32,7 +30,9 @@ development as in CI and production.
 
 You should install via homebrew
 
-     brew cask install docker
+```
+ brew cask install docker
+```
```
monfresh, line 35: There is an extra space here before `brew`.

developing/cicd/delivery-pipeline.md
rpdelaney: I think mdformat assumes you are reading the document using a markdown renderer, as is the case here (since it's hosted on github). This way is also arguably easier to write. So I get where they're coming from.

But, I'm not committed to this at all. I've played around with the consecutive numbering option and it seems like it works pretty well. It even fixes numbering errors, or at least tries.

developing/direnv/README.md
rpdelaney: That is curious. If I remove it manually, mdformat doesn't restore it. I'll follow up.

developing/command-line-tools/README.md
rpdelaney: If it bothers you, we could change this to `⬌` or `⇆` or such. Or just "and".

developing/direnv/README.md
rpdelaney: OK, got a quick dev [response](https://github.com/executablebooks/mdformat/issues/272).

It looks like there was an extra space to begin with. Have a look at [the file](https://github.com/trussworks/Engineering-Playbook/tree/882baa776df02b2f39fa09b1181570bde069def7/developing/direnv) before mdformat got to it and the extra space is there too. It's just more visible when in a fenced code block.

developing/cicd/delivery-pipeline.md
monfresh: Ah! My bad! I was looking at the rich diff inside this PR, and GitHub rendered the ordered list with the number 1 everywhere. so I assumed that was the actual output. I should have looked at the page directly from the repo, and there I see it ordered with the right numbers.

PR #200, opened 2021-10-15
Update Nix docs about Nix Channels

rogeruiz:
During an upgrade of a Nix package, we ran into a little rabbit hole and learned a lot about Nix Channels together. The Slack conversation is in a Project Slack and not Truss Slack, but we can continue having this conversation in [#code-nix](https://trussworks.slack.com/archives/C01KTH6HP7D) 🔒 to better improve our Nix documentation. 

The summary for this PR is that we may need to update certain references to commits on branches other than the default `nixpkgs-unstable` branch that is in our current documentation. Because of some of the limitations of GH Markdown, I'm using the `<details>` tag to have this documentation inline but also called out visually so folks don't lose sight of it while reading the documentation.

It's encouraged to both modify and discuss this branch.

### DIFF
--- a/developing/nix/HOWTO.md
+++ b/developing/nix/HOWTO.md
@@ -97,11 +97,12 @@ This HOWTO assumes you already have nix installed.
    fi
    ```
 
-1. Use
-   [nix-package-search](https://ahobson.github.io/nix-package-search/#/)
-   to find your package versions and add them to your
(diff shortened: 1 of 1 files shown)


rogeruiz:
Merging this as it's been over a month and I feel it's helpful and generic enough to help non-Nix users to understand how Nix consumes Git SHA Refs.

PR #203, opened 2021-10-27
ADR: Prefer TypeScript

suzubara:
This is an Engineering practice ADR that states a general preference for using TypeScript over JavaScript on projects. Please add feedback in comments, suggestions, or make your own commits to this branch!


practices/appeng/adrs/0001-use-typescript.md
```diff
@@ -0,0 +1,87 @@
...
+
+- Status: In progress
+- Deciders: @suzubara <tag yourself!>
+- Date: Oct 26, 2021
+
+## Decision drivers
+
+As consultants, we want to write understandable, maintainable code that people can use in our absence. Types provide a kind of built-in documentation for anyone looking at the code.
+
+​​TypeScript has many helpful autocomplete and snippet generation features that make development faster.
```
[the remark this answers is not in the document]
jim: I don't find development with TypeScript to be faster. Certain operations such as getting a function's arguments correctly written are slightly quicker, but I spend far more time fighting with the compiler than is saved in this way. A lot of this stems from the use of third party libraries with type definitions of varying quality/completeness, but on EASi we were careful about our dependency selection and still ran into this issue.

practices/appeng/adrs/0001-use-typescript.md
```diff
@@ -0,0 +1,105 @@
...
+- `+` Readily available linting and compile tools, including IDE integrations
+- `+` Enables many helpful autocomplete and snippet generation features that make development faster
+
+**Cons:**
+
+- `-` Adds a layer of complexity which will initially decrease productivity
+  to get on-boarded
+- `-` Potentially considerable learning overhead
+- `-` Typically compiled at build time, so run-time type errors (due to external data, user input, etc.) are not eliminated
+- `-` Requires changes to build system and config - need to be compiling types for local dev, before deploy, etc. Also need to manipulate the tsconfig. However, most widely used tools have clear docs or libraries that assist with this.
```
jim, line 60: - TypeScript is being rapidly developed and upgrading to new versions can be non-trivial as tooling across the project tends to need to be upgraded collectively.

practices/appeng/adrs/0001-use-typescript.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Use TypeScript instead of JavaScript
+
+**Pros:**
+
+- `+` Strong static typing system will help reduce possibility of bugs at
+  runtime, since type errors are captured at compile time
+- `+` Typescript has overlapping [object oriented concepts](https://levelup.gitconnected.com/typescript-object-oriented-concepts-in-a-nutshell-cb2fdeeffe6e) to a deeper level than [ES6 classes](https://www.sitepoint.com/object-oriented-javascript-deep-dive-es6-classes/) and traditional JavaScript.
+- `+` Code is clearer and easier to understand with [static typing](https://www.typescriptlang.org/docs/handbook/basic-types.html) and [interfaces](https://www.typescriptlang.org/docs/handbook/interfaces.html)
+- `+` Does not require any additional setup when used with common boilerplates such as [Create React App](https://github.com/facebook/create-react-app)
```
jim, line 47: On EASi we used CRA and still had to spend time on configuring tooling, specifically related to linters, formatters, and pre-commit checks.

suzubara:
@rswerve @jim I just fleshed out the ADR with more context, and hopefully addressed your points! please let me know if you feel like something should be added or changed.

practices/appeng/adrs/0001-use-typescript.md
```diff
@@ -0,0 +1,99 @@
...
+
+However, JavaScript has more than its share of quirks, and because it runs natively in browsers, it can be too easy to introduce bugs and errors that aren't realized until encountered by an end user. These days it is crucial to accompany JavaScript code with development tools such as linters and automated tests in order to avoid those scenarios. TypeScript is another such tool, [introduced by Microsoft](https://www.typescriptlang.org/why-create-typescript) as a superset of JavaScript that adds [type safety](https://en.wikipedia.org/wiki/Type_safety) to the language. By writing code in TypeScript instead of JavaScript, engineers can avoid entire categories of errors that would be caught well ahead of code being deployed to production.
+
+### Why type safety?
+
+The decision to use TypeScript assumes a consensus that type safety in a programming language is better than no type safety. Below are some reasons for this:
+
+- A strong static typing system can help reduce the possibility of bugs at runtime, since type errors are captured earlier, during compilation.
```
macrael, line 19: - Types free the programmer from writing an entire class of tests. You can be sure that the objects being passed into your function have fields with the correct types, many runtime checks become unneccecary and many tests cases no longer need to be written
- The most useful type check is defining wether a variable or field is required or possibly undefined. The [billion-dollar mistake](https://en.wikipedia.org/wiki/Tony_Hoare#Apologies_and_retractions) was due to an insufficienlty expressive type system

A couple suggestions. I’m sure we can drum up more. ^

jenniferlynparsons (approved):
Yay!

kylehilltruss (approved):

practices/appeng/adrs/0001-use-typescript.md
```diff
@@ -0,0 +1,107 @@
...
+
+## When to not implement this decision
+
+- If the project has no reason to use JavaScript, there may be no reason to use TypeScript. However, this decision does not preclude the option to use TypeScript as a server language if otherwise appropriate.
+- It can be very challenging and time consuming to introduce TypeScript to an existing JavaScript code base, especially if there are many dependencies and/or integrations that may or may not have TypeScript support. In this case it might not be feasible to spend time configuring TypeScript, although the option of introducing TypeScript incrementally (such as only to new code) should also be considered.
+- TypeScript can have a notoriously high learning curve, and may be too costly to adopt by certain projects with tight timing and/or staffing circumstances. However, that overhead in this situation should still be weighed carefully against the payoff of using TypeScript to prevent errors and reduce the need for writing tests later on.
```
rswerve, line 37:
```suggestion
- TypeScript can have a notoriously high learning curve, and may be too costly to adopt by certain projects with tight timing and/or staffing circumstances. However, that overhead in this situation should still be weighed carefully against the payoff of using TypeScript to prevent errors and reduce the need for writing certain kinds of tests later on.
```

practices/appeng/adrs/0001-use-typescript.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Why is this applicable to the practice as a whole
+
+As stated above, many of the projects Truss works on use JavaScript to some extent. This decision is applicable to any such project, with the intention of encouraging high code quality and consistent tooling across engineering practices.
+
+## When to not implement this decision
+
+- If the project has no reason to use JavaScript, there may be no reason to use TypeScript. However, this decision does not preclude the option to use TypeScript as a server language if otherwise appropriate.
+- It can be very challenging and time consuming to introduce TypeScript to an existing JavaScript code base, especially if there are many dependencies and/or integrations that may or may not have TypeScript support. In this case it might not be feasible to spend time configuring TypeScript, although the option of introducing TypeScript incrementally (such as only to new code) should also be considered.
```
rswerve, line 36: Worth mentioning JSDoc as a possible incremental step (or compromise solution)?

rswerve: ☝️ genuine question, by the way, not a passive-aggressive suggestion 🙂

suzubara: not at all! I am not familiar with JSDoc though, do you mind adding a suggestion?

rswerve, line 36:
```suggestion
- It can be very challenging and time consuming to introduce TypeScript to an existing JavaScript code base, especially if there are many dependencies and/or integrations that may or may not have TypeScript support. In this case it might not be feasible to spend time configuring TypeScript, although the option of introducing TypeScript incrementally (such as only to new code, or by using [JSDoc](https://www.typescriptlang.org/play#example/jsdoc-support)) should also be considered.
```

rswerve: done!

practices/appeng/adrs/0001-use-typescript.md
```diff
@@ -0,0 +1,107 @@
...
+- TypeScript implements [object-oriented concepts](https://levelup.gitconnected.com/typescript-object-oriented-concepts-in-a-nutshell-cb2fdeeffe6e) to a stronger level than [ES6 classes](https://www.sitepoint.com/object-oriented-javascript-deep-dive-es6-classes/) in vanilla JavaScript, making it a useful environment within which to learn object-oriented programming.
+
+It is also worth noting the potential pitfalls and impacts of the decision to use TypeScript:
+
+- TypeScript can add a layer of complexity for engineers unfamiliar with it, which may increase onboarding overhead.
+- TypeScript is typically compiled at build time, so run-time type errors (due to external data, user input, etc.) are not eliminated and should still be tested for.
```
gidjin, line 53: Does this fact have an effect on how easy it is to debug code while running? Ie if using the browser to step through some code will you only have access to the compiled JS or are there tools to map it to the original typescript? Mostly just asking for information, though might be good to include some details either way. However the advantage of type safety would out weigh that as a con imo.

gidjin (commented):
This is great! I left a question but I'm good approving once it's out of the WIP stage.

practices/appeng/adrs/0001-use-typescript.md
suzubara: @gidjin hmm good question! This is not an issue I've run into, I think because the difference between TS and the compiled JS code is not significant.. TS is essentially just adding type annotations, which get stripped out during compilation, without much change to the code itself.

practices/appeng/adrs/0001-use-typescript.md
suzubara: thanks!!

rswerve (approved):

PR #205, opened 2021-11-02
Distributed tracing ADR

reggieriser:
This is an ADR for using distributed tracing (with a specific recommendation of OpenTelemetry) on Truss projects.  Comments are welcome as well as direct commits to the branch.

The initial draft is sourced from a [MilMove ADR](https://github.com/transcom/mymove/blob/master/docs/adr/0061-use-opentelemetry-for-distributed-tracing.md), but reorganized to fit within the [appeng ADR template](https://github.com/trussworks/Engineering-Playbook/blob/main/practices/appeng/adrs/template.md).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,7 +8,7 @@ repos:
       - id: trailing-whitespace
 
   - repo: https://github.com/igorshubovych/markdownlint-cli
-    rev: v0.29.0
+    rev: v0.30.0
     hooks:
       - id: markdownlint
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,5 @@
 repos:
-  - repo: git://github.com/pre-commit/pre-commit-hooks
+  - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.0.1
     hooks:
       - id: check-merge-conflict
@@ -7,8 +7,8 @@ repos:
       - id: detect-private-key
       - id: trailing-whitespace
 
-  - repo: git://github.com/igorshubovych/markdownlint-cli
-    rev: v0.28.1
+  - repo: https://github.com/igorshubovych/markdownlint-cli
+    rev: v0.29.0
     hooks:
       - id: markdownlint
 
@@ -20,7 +20,7 @@ repos:
           - mdformat-gfm
           - mdformat-toc
 
-  - repo: git://github.com/trussworks/pre-commit-hooks
+  - repo: https://github.com/trussworks/pre-commit-hooks
     rev: v1.1.0
     hooks:
       - id: circleci-validate
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,7 +13,7 @@ repos:
       - id: markdownlint
 
   - repo: https://github.com/executablebooks/mdformat
-    rev: 0.7.10
+    rev: 0.7.11
     hooks:
       - id: mdformat
         additional_dependencies:
--- a/practices/appeng/adrs/0002-distributed-tracing.md
+++ b/practices/appeng/adrs/0002-distributed-tracing.md
@@ -6,7 +6,7 @@
 
 **Date Accepted:**
 
-**Reviewers:** @reggieriser \<tag yourself!>
+**Reviewers:** @jim @reggieriser @rogeruiz \<tag yourself!>
 
 ## Context
 
@@ -58,16 +58,16 @@ help answer.
 
 ### ADR Goals and Anti-goals
 
-- Goal: Choose which set of libraries to use at callsites (across programming
-  languages) within the codebase, which will be used to generate
-  distributed tracing data
+- Goal: Choose which set of libraries for generating
+  distributed tracing data to use at callsites (across programming
+  languages) within the codebase
 - Anti-goal: Committing to a specific "backend", i.e. platform or service for
   gathering, exploring, and displaying trace information
   - Sub-goal: Leaving open as many options as possible for the backend
 
 ## Decision
 
-## Use OpenTelemetry
+### Use OpenTelemetry
 
 - OpenTelemetry is an emerging industry standard
   - vendors find benefit of being in the OpenTelemetry ecosystem because they
@@ -99,13 +99,28 @@ help answer.
 
 ## Why is this Applicable to the Practice as a Whole
 
-TODO: What makes this decision generally applicable
-to Truss projects?
+Most applications will benefit from instrumentation once they are accepting
+production traffic. For many systems, the amount of work required to add
+tracing calls to the code is non-trivial and needs to be done preemptively.
+
+Projects should identify what types of calls need to be traced as a part of
+discovery and document these decisions in a project-level ADR. For example,
+a web backend might add tracing to HTTP requests, service invocations,
+database calls, and any network requests made by the backend.
 
 ## When to Not Implement This Decision
 
-TODO: What are some use cases to not implement this
-on a specific project?
+Don't use OpenTelemetry if:
+
+- the ecosystem you are working in doesn't have a mature instrumentation
+  library
+- the project already has a working instrumentation solution using a
+  proprietary hosted service that is satisfying the project's needs
+- the project is too small to support the effort required to implement
+  tracing. If this is the case, it is probably worth investigating if there
+  are other implementation options for the project, potentially in
+  higher-level languages or batteries-included frameworks.
+- it isn't possible to setup a backend to collect data
 
 ## Alternatives Considered
 
@@ -136,5 +151,6 @@ on a specific project?
 
 ## Links
 
-TODO: Link to external resources
-and project implementations of this decision.
+- [Modern Observability with OpenTelemetry](https://youtu.be/_OXYCzwFd1Y) (YouTube video, 12 minutes)
+- [Ted Young on Observability and the Release of OpenTelemetry 1.0](https://player.fm/series/the-infoq-podcast-2896265/ted-young-on-observability-and-the-release-of-opentelemetry-10) (The InfoQ Podcast, 33 minutes)
+- [How Distributed Tracing Solves One of the Worst Microservices Problems](https://petabridge.com/blog/why-use-distributed-tracing/) (article)
--- a/practices/appeng/adrs/0002-distributed-tracing.md
+++ b/practices/appeng/adrs/0002-distributed-tracing.md
@@ -58,9 +58,8 @@ help answer:
 
 ### ADR Goals and Anti-goals
 
-- Goal: Choose which set of libraries for generating
-  distributed tracing data to use at call sites (across programming
-  languages) within the codebase
+- Goal: Choose which set of libraries to use within a codebase (across
+  programming languages) for generating distributed tracing data
 - Anti-goal: Committing to a specific "backend", i.e. platform or service for
   gathering, exploring, and displaying trace information
   - Sub-goal: Leaving open as many options as possible for the backend
--- a/practices/appeng/adrs/0002-distributed-tracing.md
+++ b/practices/appeng/adrs/0002-distributed-tracing.md
@@ -1,6 +1,6 @@
 # 0002 - Use OpenTelemetry to instrument code for distributed tracing
 
-**Status:** In Progress
+**Status:** Accepted
 
 **Date Accepted:** 12/8/2021
 
--- a/practices/appeng/adrs/0002-distributed-tracing.md
+++ b/practices/appeng/adrs/0002-distributed-tracing.md
@@ -1,0 +1,140 @@
+# 0002 - Use OpenTelemetry to instrument code for distributed tracing
+
+**Status:** In Progress
+
+**Date Superseded:** (if applicable)
+
+**Date Accepted:**
+
+**Reviewers:** @reggieriser \<tag yourself!>
+
+## Context
+
+### Objective
+
+_Give developers and operators an easier way to understand the behavior and
+structure of running systems by instrumenting code for distributed tracing._
+
+### Why Distributed Tracing?
+
+> Observability is defined as the ability of the internal states of a system to
+> be determined by its external outputs.
+
+There are several established techniques for gaining observability into complex
+software systems.
+
+- Logging
+  - helpful for known knowns
+  - e.g. "I want to know when a specific condition is reached"
+- Metrics
+  - helpful for assessing known unknowns
+  - e.g. "How many requests per second was the system handling last Tuesday?"
+- Distributed Tracing
+  - helpful for learning about unknown unknowns
+  - e.g. "What was the execution context for User X that caused their
+    interaction to timeout last Tuesday?"
+
+Some of the benefits of distributed tracing, as outlined in
+[this](https://petabridge.com/blog/why-use-distributed-tracing/) article are:
+
+- radically improves developer productivity and output
+- works across multiple applications, programming languages, and transports
+- improve time to market
+- facilitates excellent cross-team communication and cooperation
+
+Here are several example scenarios or questions that distributed tracing can
+help answer.
+
+- As a new engineer on the team, I want to understand how many separate systems
+  are involved when a certain user type logs in and the first page is rendered.
+- As an operations engineer, I want to know how many SQL queries are executed
+  for a given endpoint or interaction.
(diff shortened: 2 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -1,14 +1,14 @@
...
     hooks:
       - id: check-merge-conflict
       - id: check-yaml
       - id: detect-private-key
       - id: trailing-whitespace
 
-  - repo: git://github.com/igorshubovych/markdownlint-cli
-    rev: v0.28.1
+  - repo: https://github.com/igorshubovych/markdownlint-cli
+    rev: v0.29.0
```
reggieriser, line 11: I ran into problems trying to run pre-commit with the `git://` URL:

```
fatal: remote error: 
  The unauthenticated git protocol on port 9418 is no longer supported.
Please see https://github.blog/2021-09-01-improving-git-protocol-security-github/ for more information.
```

The error message pointed to [this article](https://github.blog/2021-09-01-improving-git-protocol-security-github/).  It seemed we needed to migrate over to `https://` instead, so I did that.  I also updated to the latest version of these linters.

practices/appeng/adrs/0002-distributed-tracing.md
```diff
@@ -0,0 +1,140 @@
...
+  - some OpenTelemetry libraries and tools may trail their vendor-supported
+    counterparts
+  - instrumentation for tracing may be a vector for performance overhead
+
+## Why is this Applicable to the Practice as a Whole
+
+TODO: What makes this decision generally applicable
+to Truss projects?
+
+## When to Not Implement This Decision
```
jim, line 105: * Don't use OpenTelemetry if the ecosystem you are working in doesn't have a mature library

practices/appeng/adrs/0002-distributed-tracing.md
```diff
@@ -0,0 +1,140 @@
...
+- Cons
+  - as an abstraction layer, OpenTelemetry may prohibit usage of vendor-
+    specific capabilities
+  - some OpenTelemetry libraries and tools may trail their vendor-supported
+    counterparts
+  - instrumentation for tracing may be a vector for performance overhead
+
+## Why is this Applicable to the Practice as a Whole
+
+TODO: What makes this decision generally applicable
```
jim, line 102: Most applications will benefit from instrumentation once they are accepting production traffic, and for many systems, the amount of work required to add tracing calls to the code is non-trivial and needs to be done preemptively.

Projects should identify what types of calls need to be traced as a part of discovery and document these decisions in a project-level ADR. For example, a web backend might add tracing to HTTP requests, service invocations, database calls, and any network requests made by the backend.

practices/appeng/adrs/0002-distributed-tracing.md
rogeruiz: I like this criteria. My understanding is that it's possible that this could change over time if a contract scales and the project gets bigger. 

Is it a good idea to include a better breakdown of what the thresholds look like for implementing tracing?

It can be difficult to add tracing after the fact for a mature project. How can this ADR address having an MVP that is friendly to building out OpenTelemetry if the project scales up and eventually needs it?

I speak from experience on MilMove that implementing OpenTelemetry came at the cost of a large number of PRs that were difficult to review.

I'm musing on wording here that explains how we can possibly architect a solution that is OpenTelemetry-friendly even though it hits the points you made of not needing to implement it.

practices/appeng/adrs/0002-distributed-tracing.md
rogeruiz: LOL same brain!

I added some thoughts on this in the [other comment here](https://github.com/trussworks/Engineering-Playbook/pull/205#discussion_r741496096) but this definitely seems like the more appropriate space for it. 

Given this, I'd say that it might be interesting (currently) to assume that we shouldn't assume that OpenTelemetry or application tracing is something we would suggest against given how costly it can be to do it in the future.

Considering the future of web applications, it seems like Observability is something that is here to stay. Maybe this means we need to consider that in this ADR.

practices/appeng/adrs/0002-distributed-tracing.md
mikena-truss: Just a couple points here:

- This ADR doesn't supersede the need for implementation documentation, which should have more of this information
- Hopefully the standardization of this makes it easier for all projects (including small ones) to quickly implement, since we have the in house knowledge

practices/appeng/adrs/0002-distributed-tracing.md
```diff
@@ -0,0 +1,156 @@
...
+
+**Date Accepted:**
+
+**Reviewers:** @jim @reggieriser @rogeruiz \<tag yourself!>
+
+## Context
+
+### Objective
+
+_Give developers and operators an easier way to understand the behavior and
```
mikena-truss, line 15: One small trick I like to try in these short synopses is tying this back to the client and/or user. It's ok if that's not 100% possible (ex. if something made our developers 1000% happier at low implementation cost with little/no benefit to the user, we'd just do it), but it's helpful practice for ensuring we're creating clear value. For example, we want to keep/recover service health and allow users to consistently use our applications.

practices/appeng/adrs/0002-distributed-tracing.md
```diff
@@ -0,0 +1,156 @@
...
+- Metrics
+  - helpful for assessing known unknowns
+  - e.g. "How many requests per second was the system handling last Tuesday?"
+- Distributed Tracing
+  - helpful for learning about unknown unknowns
+  - e.g. "What was the execution context for User X that caused their
+    interaction to timeout last Tuesday?"
+
+Some of the benefits of distributed tracing, as outlined in
+[this](https://petabridge.com/blog/why-use-distributed-tracing/) article are:
```
felipe-lee, line 38: Small nitpick but it's generally good to include more descriptive text in links for accessibility. See this [A11y Project post on accessible links](https://www.a11yproject.com/posts/2019-02-15-creating-valid-and-accessible-links/).

practices/appeng/adrs/0002-distributed-tracing.md
```diff
@@ -0,0 +1,156 @@
...
+  certain cohort of users on a regular basis.
+- As an engineer, I want to prove that an optimization I wrote is effective
+  in a production environment.
+- As a load tester, after I have shown that a problem exists, I want to
+  understand how the system is interacting so I can debug and fix the issue.
+
+### ADR Goals and Anti-goals
+
+- Goal: Choose which set of libraries for generating
+  distributed tracing data to use at callsites (across programming
```
felipe-lee, line 62: What does callsites mean? Like does this mean the places within the code being instrumented? E.g. middleware, functions, etc.?

felipe-lee (commented):
Just two small things, but otherwise looks pretty good, thanks for putting this together.

practices/appeng/adrs/0002-distributed-tracing.md
reggieriser: I changed the objective slightly to mention a more consistent experience for the user.

practices/appeng/adrs/0002-distributed-tracing.md
reggieriser: Good point.  I changed this to reference the title of the article.

practices/appeng/adrs/0002-distributed-tracing.md
reggieriser: I think it just means anywhere code is being called (this was language that came from the sourced MilMove PR).  Here's how Wikipedia [defines a call site](https://en.wikipedia.org/wiki/Call_site).  I'm open to alternative language here if you have a suggestion.

practices/appeng/adrs/0002-distributed-tracing.md
```diff
@@ -0,0 +1,156 @@
...
+Don't use OpenTelemetry if:
+
+- the ecosystem you are working in doesn't have a mature instrumentation
+  library
+- the project already has a working instrumentation solution using a
+  proprietary hosted service that is satisfying the project's needs
+- the project is too small to support the effort required to implement
+  tracing. If this is the case, it is probably worth investigating if there
+  are other implementation options for the project, potentially in
+  higher-level languages or batteries-included frameworks.
```
gidjin, line 122: Above the ADR says a project gets the most benefit once accepting production traffic, but that adding tracing is non trivial and should be done preemptively. Is the recommendation to add it at the beginning or once a project is mature enough to be in production?

Often our projects start off as small prototypes that then grow into a larger application. Is the recommendation for a higher-level language or batteries-included framework applicable to those cases? Is there a recommendation on when to consider adding instrumentation for maximum benefit when starting with a small prototype that eventually needs to transition to a large production application?

rogeruiz: when I read this, I felt that _the recommendation is to add it at the beginning of a project_ is what I interpreted this as. But I think some clarification would be super helpful.

reggieriser: Thanks for the feedback.  I added in some additional wording to specify that this work normally needs to occur in the early stages of the project.  I think the "small project" exclusion is really for something we don't anticipate will ever turn into something larger, not as a way to defer thinking about instrumentation until just before production.

practices/appeng/adrs/0002-distributed-tracing.md
felipe-lee: I guess I would maybe suggest removing the call site bit since imo it doesn't add much. Like to me, something like

> Choose which set of libraries to use within a codebase (across programming languages) for generating distributed tracing data. 

gets the same point across, but maybe I'm missing something.

.pre-commit-config.yaml
reggieriser: The main "git://" vs "https://" issue was fixed upstream in the main branch (just merged it in), but I did include a couple of minor version updates since then.

practices/appeng/adrs/0002-distributed-tracing.md
```diff
@@ -0,0 +1,157 @@
...
+    development environment
+  - e.g. can use Honeycomb.io in the experimental commercial-cloud hosted
+    environment
+  - e.g. can swap in AWS X-Ray for use in GovCloud hosted environments
+- Cons
+  - as an abstraction layer, OpenTelemetry may prohibit usage of vendor-specific
+    capabilities
+  - some OpenTelemetry libraries and tools may trail their vendor-supported
+    counterparts
+  - instrumentation for tracing may be a vector for performance overhead
```
rogeruiz, line 97: I am not sure how I feel about this line here. I don't know if it's accurate or not. My understanding of observability / OpenTel is that the metrics collected and shipped would actually help with performance overhead and not contribute to it.

Is this line suggesting that adding instrumentation could cause the app to slow down? Or is it suggestion that developer performance could slow down due to adding instrumentation?

```suggestion
  - instrumentation for tracing may be a vector for developer-performance overhead because of additional work to implement into roadmap and product planning
```

rogeruiz (approved):
lgtm 🌈 thanks @reggieriser. This is a huge step forward imo. I appreciate all the additional changes you took in as well from everyone else.

I have one suggestion, but [I see that the original ADR also used the same language](https://github.com/transcom/mymove/blob/master/docs/adr/0061-use-opentelemetry-for-distributed-tracing.md). I think it'd be better to remove the ambiguity but I'll leave that to you.

Also I'll add that the **date accepted** will need to be updated before merging right? Or is that done after the fact?

jim (approved):

gidjin (approved):

practices/appeng/adrs/0002-distributed-tracing.md
```diff
@@ -0,0 +1,157 @@
...
+- OpenTelemetry has well-maintained libraries for the languages used in the
+  layers of most Truss projects
+  - i.e. Go (back-end); JavaScript (front-end); Python (load testing); etc
+- Easily swappable back-ends
+  - e.g. could choose a local Docker version of OpenZipkin for an all-local
+    development environment
+  - e.g. can use Honeycomb.io in the experimental commercial-cloud hosted
+    environment
+  - e.g. can swap in AWS X-Ray for use in GovCloud hosted environments
+- Cons
```
ahobson, line 92: The [opentelemetry-go](https://github.com/open-telemetry/opentelemetry-go) implementation is, unfortunately, not stable. Metrics are only in alpha, so the API can (and has) changed. I think it's still probably the right choice, but it is a con.

ahobson, line 92: The [opentelemetry-go-contrib/instrumentation](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation) is not merging pull requests. We've had to create a fork of the [http tracing/metrics](https://github.com/trussworks/otelhttp) part because it doesn't have all the features we need. For go, it seems like we're going to have to maintain the metrics library ourselves.

ahobson:
Thank you! This is really great progress!

practices/appeng/adrs/0002-distributed-tracing.md
felipe-lee: I think it may be in reference to the fact that in order to run otel, we are in fact running another service and the main application will need to be passing metrics, traces, etc. to the collector so that can impact performance. There's things we can do to modify how much it affects, e.g. change what percentage of things get sampled, but regardless, there will be some impact.

felipe-lee (approved):
LGTM, thanks for putting this together and making updates!

practices/appeng/adrs/0002-distributed-tracing.md
reggieriser: My interpretation matched what @felipe-lee said.

practices/appeng/adrs/0002-distributed-tracing.md
reggieriser: Thanks for the context on the challenges with the go implementation.  I think I'd rather that type of info go outside the ADR in a "best practices" type section rather than bake it into the ADR.  I'm hopeful that it gets better soon.

ahobson (approved):

practices/appeng/adrs/0002-distributed-tracing.md
rogeruiz: I don't really agree here. The OpenTel specification addresses this. I think it's too general to say that it could impact performance. The client and SDK authors take this into account. I think it's up to the users of these tools to verify, but not be concerned with the performance issues.

[=> OpenTel Spec on performance](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/performance.md)

reggieriser: I suggest we leave this as is for now -- I think there's always the potential that adding another service/dependency/network hop could affect performance, particularly if not integrated or configured properly.  Perhaps we can cover specifics about any performance impacts (or lack of) in the playbook in a "best practices" section (which I expect we will be building as a follow-on to this ADR).

rogeruiz: I don't feel like what I'm trying to say is coming across appropriately. But I'll try again. The performance that this is implying is that it's affecting the application. I'm not sure that's actually the original concern here. The performance could very well happen on the OpenTel side, the collector, the parser, the aggregator of the data. But the ADR is clear that without a backend the OpenTel calls would be _lightweight_ and a _noop_ which to me are the opposite of _a vector for performance overhead_. 

But like I said in my original review, this is not a blocker, just wanted to make sure it's captured and expressed somewhere. 👍🏽 I wasn't around for the original ~ADR~ MDR being written, but I would have had the same concern then too. The spec is very clear that this is a concern of the authors of OpenTel tools and not the consumers of those tools who are being targeted with this MDR/ADR.

practices/appeng/adrs/0002-distributed-tracing.md
rogeruiz: I think the instability of some implementations as a Con I think is a valid thing to attribute in this ADR. One of the Pros in this document are that _OpenTelemetry has well-maintained libraries for the languages used in the
   layers of most Truss projects_ which is opposite to the experience @ahobson mentioned. Maybe a good Con to add is that the specification is still early with pre-1.0 versions of libraries being used.

macrael (approved):
Thanks @reggieriser

practices/appeng/adrs/0002-distributed-tracing.md
reggieriser: I went ahead and added a note about pre-1.0 libraries as a potential con.

PR #209, opened 2021-12-01
Remove deprecated nix install flags

mboren:
--darwin-use-unencrypted-nix-store-volume is deprecated
and
--no-daemon is no longer supported and prevents the installer from running

Here's the output I got when trying to use these flags to install on a new laptop:
```
marty@martys-MacBook-Pro ~ % sh <(curl -L https://nixos.org/nix/install) --darwin-use-unencrypted-nix-store-volume --no-daemon
<... curl stuff omitted ...>
Warning: the flag --darwin-use-unencrypted-nix-store-volume
         is no longer needed and will be removed in the future.

Error: --no-daemon installs are no-longer supported on Darwin/macOS!
```

### DIFF
--- a/developing/configuration/README.md
+++ b/developing/configuration/README.md
@@ -3,15 +3,19 @@
 Projects should adhere to the following guidelines for managing configuration:
 
 1. Configuration values should be provided to programs via environment variables.
-2. Programs should validate configuration immediately when they are run.
-3. If config is incomplete or invalid, programs should:
+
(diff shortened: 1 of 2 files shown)


rogeruiz:
@mboren I think you need to run the `pre-commit` commands locally to figure out what the issue is here. The content looks good, but without CI passing I can't approve it as is.

rogeruiz (commented):
lgtm 🌈 thanks for doing this @mboren

PR #210, opened 2021-12-11
Update eng playbook to reflect per-user install of nix

gidjin:
Installing nix on my new laptop I noticed that `per-user` is the default installation now and this PR is to update the recommendations in this playbook to reflect that.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -2,7 +2,11 @@
 
 # if nix is installed, use it
 if [ ! -r .nix-disable  ] && has nix-env; then
-    export NIX_PROFILE=/nix/var/nix/profiles/eng-playbook
+    if [ -w "/nix/var/nix/profiles/per-user/${LOGNAME}" ]; then
+      export NIX_PROFILE="/nix/var/nix/profiles/per-user/${LOGNAME}/eng-playbook"
+    else
+      log_error "Cannot determine writable location for your NIX_PROFILE"
+    fi
     PATH_add ${NIX_PROFILE}/bin
     nix_file=nix/default.nix
 
--- a/.envrc
+++ b/.envrc
@@ -2,11 +2,7 @@
 
 # if nix is installed, use it
 if [ ! -r .nix-disable  ] && has nix-env; then
-    if [ -w "/nix/var/nix/profiles/per-user/${LOGNAME}" ]; then
-      export NIX_PROFILE="/nix/var/nix/profiles/per-user/${LOGNAME}/eng-playbook"
-    else
-      log_error "Cannot determine writable location for your NIX_PROFILE"
-    fi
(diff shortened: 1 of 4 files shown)


nix/update.sh
```diff
@@ -2,14 +2,17 @@
 
 set -euo pipefail
 
-if [ ! -v NIX_PROFILE ]; then
+if [ -z "$NIX_PROFILE" ]; then
```
ahobson, line 5: `-v` seems to be a modern bashism so maybe we shouldn't rely on it. We'd have to use `[ -z "${NIX_PROFILE+x}" ]` so the undefined variable setting above doesn't bite us

.envrc
```diff
@@ -2,7 +2,11 @@
 
 # if nix is installed, use it
 if [ ! -r .nix-disable  ] && has nix-env; then
-    export NIX_PROFILE=/nix/var/nix/profiles/eng-playbook
+    if [ -w "/nix/var/nix/profiles/per-user/${LOGNAME}" ]; then
+      export NIX_PROFILE="/nix/var/nix/profiles/per-user/${LOGNAME}/eng-playbook"
+    else
+      log_error "Cannot determine writable location for your NIX_PROFILE"
+    fi
```
ahobson, line 9: I know I proposed this pattern, but upon further reflection, maybe it would be better to just set the `NIX_PROFILE` and then have nix fail if it's not right. That way we'd probably get better error messages.

developing/nix/HOWTO.md
```diff
@@ -44,14 +44,16 @@ up-to-date.
 
    set -euo pipefail
 
-   if [ ! -v NIX_PROFILE ]; then
+   if [ -z "$NIX_PROFILE" ]; then
```
ahobson, line 47: Same undefined variable problem as below

nix/update.sh
gidjin: Ah I was wondering why this started failing, I am on a new mac and it has `bash 3.2` need to upgrade my bash apparently. `-v` was throwing errors for me, hence the change. I'll update bash and see if that resolves it.

.envrc
gidjin: I can go either way to be honest, but I updated the PR to go with your recommendation.

ahobson: As I proposed, the user will get an error, but it won't tell them which path the script tried to use, which is less than helpful.

rogeruiz: [I see the `HOWTO.md` got updated](https://github.com/trussworks/Engineering-Playbook/pull/210/commits/6c3f160ad94362ed39f86761a30fab46adde1820), but this must also get updated too since this is where the error would occur :vs: the example for an `.envrc` file.

rogeruiz (approved):
lgtm 🌈

ahobson (approved):

rogeruiz (commented):
lgtm 🌈

ship it!

PR #211, opened 2021-12-20
Add decision guide with engineering values

mikena-truss:
First draft at adding engineering values for making decisions.

Here was my take:
* I combined common themes from the [ADR Value mapping exercise](https://miro.com/app/board/o9J_liBoAmk=/) with the [Team Charter](https://github.com/trussworks/Engineering-Playbook/blob/main/practices/appeng/team-charter.md)
* I set up an example using our most common user groups
* I guided through how to apply those values

Feedback that would helpful:
* Do these values seem like a good starting point for us? My hunch is that could be tangled and twisted in a lot of ways, but these align with my understanding of our practice atm.
* Is this digestible/usable?
* Is there any glaringly missing content?


practices/appeng/decision-guide.md
```diff
@@ -0,0 +1,200 @@
...
+our software is underneath the hood,
+if a user can't access it
+and perform its intended purpose with ease,
+we have not built good software.
+This includes making our software accessible
+to our users,
+stakeholders,
+and team.
+
+> ... Postgres internal types align well with common API level types
```
mikena-truss, line 100: Open to a better example here.

reggieriser: I really like the Postgres examples in the doc, although agree that this one feels a little forced and is more specific to public API consumers.  I'm also not sure what types Postgres has that give it a significant advantage in this respect over other major relational databases.  That said, I'm not sure I have a great example either, but I'll toss out a few thoughts:

- Performance-related examples?  Not sure if Postgres has significant advantages in that realm, but seems like something that could impact an end user.
- Lower (or non-existent) licensing/support costs?  Other commercial databases like Oracle can cost a fortune for a license as well as ongoing support.  End users may not directly care about this, but it can lower the cost of building/maintaining software.
- More security transparency being open source?  That may be a stretch too, but end users might indirectly benefit from that as well.

ferlatte: Better defaults/fewer traps, maybe? Postgres enforces stricter SQL conformance than MySQL does by default, which we believe means fewer errors in production; as I recall it also defaults to better settings for character sets, collation, and time/date fields.

Postgres does let you define your own column level types in ways that MySQL doesn't, although I'm not sure if that's necessarily a compelling argument here (although, PostGIS is *amazing*).

PR #232, opened 2022-03-15
Update pre-commit cache key generation

jsclarridge:
Update the method of generating the pre-commit cache key. That was my mistake for not catching this before. `checksum` is from CircleCI; `hashFiles` is the GitHub Actions expression.

### DIFF
--- a/.github/workflows/pre_commit.yaml
+++ b/.github/workflows/pre_commit.yaml
@@ -17,7 +17,7 @@ jobs:
       - name: cache
         uses: actions/cache@v2
         with:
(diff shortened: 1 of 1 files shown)


rpdelaney (approved):
How do we know if it's working or not?

jsclarridge:
> How do we know if it's working or not?

In the GitHub Action workflow, under the `cache` step, compare the generated cache key name for this PR to an earlier one.

rpdelaney:
I mean, how do we know if it's caching correctly? I guess it's this?

Screen Shot 2022-03-15 at 16 05 25

PR #234, opened 2022-03-18
ADR for PostgreSQL as a relational store

reggieriser:
This is the initial draft for a practice-wide ADR to document the decision to use PostgreSQL as our preferred relational DBMS.


practices/appeng/adrs/0005-postgresql-as-relational-store.md
```diff
@@ -0,0 +1,87 @@
...
+
+- `+` Mature DBMS with long history
+- `+` Traditional paid software backed by a company and a support team
+- `+` Supported by most database tools and ORMs
+
+#### Cons
+
+- `-` Licensing fees and support can be very expensive
+- `-` Not open source so you are reliant on company for fixes
+- `-` Support by ORMs can be spotty
```
sandy-wright, line 83: I do have some experience with Oracle databases, and one con you could consider adding is how difficult it is to search for the syntax of Oracle SQL vs. MySQL or other more popular dialects. "SQL" is everywhere, so any version that's markedly less popular than the others is liable to be drowned out of the results when you're looking for help.

felipe-lee (commented):
Some things I brought up at the server guild but adding here too:

* Should there be info on cloud provider support, especially AWS since we use that for many projects? (Reggie mentioned that that's what `reuse of infrasec practices` was calling out, but maybe would expand on that).
* Should we call out that some tutorials/frameworks might use `sqlite` for tests, even if your deployment/dev DB is a different thing? (Reggie mentioned that it's probably bad practice to do that, and I agree, but wondering if that should be here, or if that belongs more in like a best practices doc).
* In some project ADRs we'll sometimes mention as a pro/con that people on the project have/lack experience with certain tools. I'm not sure that really belongs in this one since it's a Truss-wide ADR and we probably have vastly differently experience levels across Truss for each of these. Though, maybe it should be called out that we've use postgres on most projects that need a relational DB (at least I think we do).
* Since we use AWS, I don't know if it's worth it to call out Aurora Postgres. Like I don't know if that's different from what we use in RDS tbh. I _think_ it is, but could definitely be wrong. And _if_ it is different, maybe it just falls under the "Commercial" group (though pro/cons would really be more a mix of that group and the standalone Postgres section)

practices/appeng/adrs/0005-postgresql-as-relational-store.md
felipe-lee: And, I think you also brought this up in the server guild meeting Sandy, but also error codes can a pain to figure out with Oracle.

ferlatte:
@felipe-lee Aurora Postgres is probably worth mentioning as a "here's a thing to consider if you're building a really big or HA site". For most of our projects, I think it's probably overkill and last I checked it was more expensive (but that should be validated, since things change with AWS pricing).

felipe-lee:
> @felipe-lee Aurora Postgres is probably worth mentioning as a "here's a thing to consider if you're building a really big or HA site". For most of our projects, I think it's probably overkill and last I checked it was more expensive (but that should be validated, since things change with AWS pricing).

Ah, good to know about the pricing. I guess given that it can change, that can be one of the bullet points even, "compare price at the time and see which might be more worth it"?

reggieriser:
Thanks for all the feedback!  I just pushed up changes that I think cover most of the points above.  I didn't include anything around Felipe's mention of using two different databases in different environments -- I think that may be more suitable for a best practices doc if we want to address it.

practices/appeng/adrs/0005-postgresql-as-relational-store.md
```diff
@@ -0,0 +1,95 @@
...
+- `-` Differences in SQL syntax and error codes can make it difficult to get help compared to more popular
+  open-source options
+
+## Links
+
+TBD
```
felipe-lee, line 95: **suggestion (non-blocking)**: I'm not sure if there's any links you want to add, but if not, can we get rid of this section?

felipe-lee (approved):
Awesome, thanks for making those updates. LGTM, I had just one comment, but it's a small thing.

practices/appeng/adrs/0005-postgresql-as-relational-store.md
reggieriser: Thanks for the review/approval!  I'll make sure I add links or remove the `Links` section before merging.  I also need to do some other cleanup to the header (status change, date accepted, etc.) before merging too.  I'll take it out of draft tomorrow and put a call out for any additional comments (probably should wait a few days after that before I merge).

felipe-lee: Sounds good, thanks!

practices/appeng/adrs/0005-postgresql-as-relational-store.md
```diff
@@ -0,0 +1,95 @@
...
+Historically, most Truss projects deploy a database on AWS. Note that AWS provides two PostgreSQL options:
+RDS PostgreSQL and Aurora PostgreSQL. The typical Truss project will likely use RDS PostgreSQL since Aurora targets
+very large or high-availability sites. That said, evaluate the features and pricing for each against your project's
+needs.
+
+## When to Not Implement This Decision
+
+- If the project's client/sponsor has an established and/or inflexible existing DBMS requirement
+- If the nature of the data lends itself to a different type of data store (NoSQL database, graph database, etc.)
+- The project is so small, self-contained, or temporary that a lightweight solution like SQLite is more appropriate
```
mikena-truss, line 40: Curious if this is relevant. Maybe personal bias, but postgres has become accessible enough to run where the migration from sqlite to postgres outweighs the ease of using sqlite

mikena-truss (approved):
lgtm 👍

practices/appeng/adrs/0005-postgresql-as-relational-store.md
reggieriser: I guess I was just hedging a bit for cases (like an example, starter, or template project) where being able to spin something up with a file-based database like SQLite was helpful in removing a little bit of setup friction.  I don't anticipate that we'd ever use SQLite in a deployed project.

JasonLin0991 (approved):

PR #259, opened 2022-04-08
Add Docusaurus message

mikena-truss:
Add a disclaimer/alert for Docusaurus changes to main README

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
-⚠️❗ __The deployed version of the engineering playbook is now on [Docusaurus](https://playbook.truss.dev/).
+⚠️❗ __The engineering playbook is now deployed at [playbook.truss.dev](https://playbook.truss.dev/) using [Docusaurus](https://docusaurus.io/).
+We recommend you read it there.
 Links and formatting from the old layout may be broken.
 Known problems are reported to [GitHub issues](https://github.com/trussworks/Engineering-Playbook/issues).
 Feel free to open a PR if you see something that can be fixed.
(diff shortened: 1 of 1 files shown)


rpdelaney (approved):
One non-blocking comment. Thank you for taking the initiative on this!

README.md
```diff
@@ -1,3 +1,9 @@
+⚠️❗ __The deployed version of the engineering playbook is now on [Docusaurus](https://playbook.truss.dev/).
```
rpdelaney, line 1: I sincerely don't know if this thought is helpful or just pedantic, so I leave it to others to decide :)

I think Docusaurus is just a markdown-to-html rendering engine, so the eng playbook is not really "on" Docusaurus?

```suggestion
⚠️❗ __The engineering playbook is now deployed on [playbook.truss.dev](https://playbook.truss.dev/) and it should be read there.
```

mikena-truss:
```suggestion
⚠️❗ __The deployed version of the engineering playbook is now deployed at [playbook.truss.dev](https://playbook.truss.dev/) using [Docusaurus](https://docusaurus.io/).
We recommend you read it there.
```

@rpdelaney how about this?

rogeruiz: I'll say that we saw this same issue with MilMove when discussing documentation. We eventually stopped calling it Docusuarus as not only is it sometimes difficult to say / spell check. It was also causing confusion as Docusaurus is a Facebook project and was causing folks to get further confused. I like the second suggestion here @mikena-truss

rpdelaney: Looks good!

PR #271, opened 2022-05-31
A JavaScript dependency management ADR

rogeruiz:
[More of a description to come. For now you can follow along in the Slack thread on this](https://trussworks.slack.com/archives/CD7MJLFJ4/p1653663891064749). 🔒 

## Viewing this PR locally

Checkout this branch and then run `yarn start`. You may need to run `yarn install` if it's your first time checking out the repository since the migration to Docusaurus. Once the server is running, you should be able to access the ADR directly at this URL: [http://localhost:4000/docs/practices/appeng/adrs/javascript-dependency-management](http://localhost:4000/docs/practices/appeng/adrs/javascript-dependency-management).


docs/practices/appeng/adrs/0005-javascript-dependency-management.mdx
```diff
@@ -0,0 +1,273 @@
...
+
+Truss recommends that projects use _NPM_ instead of _Yarn_ as it is included with
+Node and has commercial support. Projects that leverage _Yarn_ or any other
+**JavaScript Dependency Management Tool** (JDMT from now on) will need to
+document this in ways outside of the scope of this ADR.
+
+This ADR is not meant to endorse one JDMT over the other and ultimately places the
+decision on engineering teams using these tools. There are certain ways these
+tools work together and this document will aim to cover those scenarios
+equitably.
```
rogeruiz, line 50: I'm not too sure about this section here. I think it might be good to establish use of _NPM_ over _Yarn_ for Truss projects, but I also have not been on any Truss projects that use _NPM_ over _Yarn_. This is why this opening paragraph is worded so weirdly. 

[=> Slack thread on this topic](https://trussworks.slack.com/archives/CD7MJLFJ4/p1654609942295709).

docs/practices/appeng/adrs/0005-javascript-dependency-management.mdx
```diff
@@ -0,0 +1,273 @@
...
+
+You will notice that all of the packages listed are set to specific versions
+(instead of a range using the `^` or `~` characters). This is to help ensure
+that installing dependencies is deterministic and it also means we are in full
+control of all version updates.
+
+By default, JDMT will use the `^` caret character for the version which does not
+lead to control of what version get installed when running the JDMT installation
+command. Avoid using semantic version ranges when install dependencies by using
+some of the commands listed above around installation.
```
rogeruiz, line 147: For this section, I think we should have a recommendation for using `.npmrc` file to set the `--save-exact` flag to be true for _NPM_. This way we don't need to rely on folks having the correct flags when running `npm install` commands. I don't believe there is an equivalent _Yarn_ file, but I will look after posting this.

[=> See docs here](https://docs.npmjs.com/cli/v8/using-npm/config)

rogeruiz: Nothing to configure _Yarn_ to leverage the same kind of behavior as _NPM_ here. But there is a `.yarnrc` file that exists.

[=> _Yarn_ docs on configuration](https://classic.yarnpkg.com/en/docs/yarnrc)

docs/practices/appeng/adrs/0005-javascript-dependency-management.mdx
```diff
@@ -0,0 +1,273 @@
...
+
+Ultimately, the goal is to minimize risk by ensuring that whatever changes you
+do make can be easily tested, and if an issue is found it's straightforward to
+either debug or roll back from. That all comes down to identifying what
+package introduced the issue, and how to resolve it. If a change involves
+updating many packages _and_ making many code changes at once, that will be
+much more difficult.
+
+### Working with files used and generated by JDMT
+
```
rogeruiz, line 250: This section will discuss using `.gitattributes` and `.gitignore` to prevent editing/diffing Lock files and prevent mixing those Lock files respectively. I will probably get to this at some point today or tomorrow.

gidjin (approved):
LGTM, this feels pretty close to the current project I'm  on. I like the use of tabs to show NPM or Yarn commands.

PR #274, opened 2022-06-30
Add additional troubleshooting step for yubikey tutorial

gidjin:


esacteksab (approved):

ahobson (changes requested):
Sorry, I think we need to support both Intel and M1 macs for a bit longer in our docs

ahobson (approved):

rogeruiz:
I don't think we need to fix things in this PR, but I'm noticing a lot of broken images on this page and also would suggest that this page gets broken down into sub-pages as it's very long at the moment.

PR #279, opened 2022-07-08
Revise Ruby documentation to recommend asdf

jim:
Based on [a Slack discussion](https://trussworks.slack.com/archives/CDJ9JAD26/p1657075184688319), we're now recommending that folks use asdf to manage Ruby.

I tested the steps locally, but without a fresh macOS installation, it's hard to be sure that there aren't any gaps to trip over. These guide revisions would benefit from someone who doesn't use Ruby already following along and looking for issues.

PR #281, opened 2022-07-13
Package manager spike

jim:
Spawned a bit by #279, I broke the asdf-specific docs out into a new page and added a draft of some top-level guidance for how we might think about package managers. [The new content is a little easier to read by browsing the branch](https://github.com/trussworks/Engineering-Playbook/tree/jb-package-manager-spike/docs/developing/package-managers), but note that the [admonitions](https://docusaurus.io/docs/markdown-features/admonitions) don't render regardless.

Please add your thoughts, concerns, etc. as a code review!


docs/developing/package-managers/README.md
```diff
@@ -0,0 +1,48 @@
...
+Use [asdf](package-managers/asdf) to install and activate specific versions of language runtimes. You will need to also setup the corresponding [language plugin](https://asdf-vm.com/manage/plugins.html) for the language you want to manage with asdf.
+
+## Language-specific Dependency Managers
+
+Many programming languages have their own package managers to install and manage project dependencies. We recommend using the following dependency managers:
+
+- [npm](https://www.npmjs.com/) for JavaScript and TypeScript (Node)
+- [Poetry](https://python-poetry.org/) for Python
+- [RubyGems](https://rubygems.org/) for Ruby
+- `go mod` for Go
```
rpdelaney, line 30: I think you do a good job of explaining the distinction between a package manager and a dependency manager. Since this page is about package managers, it could be argued that we shouldn't have recommendations about dependency managers here as those are truly language-specific tools.

docs/developing/package-managers/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Use Docker for Services and Anything Else
+
+If a project needs to connect to a running service (such as a database), we recommend using Docker Compose to orchestrate and manage those dependencies without needing to install anything directly on a developer's machine.
+
+Additionally, Docker is a useful way to run specific versions of software that doesn't share an ecosystem with the rest of a project. We've done this successfully to leverage [Flyway](https://flywaydb.org), a Java-based database migration system, within a project that didn't otherwise use Java.
+
+## Other Project-specific Software
+
+Any software that doesn't described by one of the above categories belongs to this one. It's difficult to manage this kind of software because it tends to not neatly fit within the domain of one of the tools covered in earlier sections.
```
ferlatte, line 40: Do we have any examples from existing projects that fall into this category?

docs/developing/package-managers/README.md
esacteksab: As someone who doesn't understand the difference between npm, yarn, npx, pnpm, I would love a page like this.

rogeruiz: I don't know if we need to split up the difference into different pages of documentation. They're still managing dependencies and packaged software regardless of it's based on the chip architecture, operating system, or programming language. I think there's a good breakdown of what these differences are and what these tools are. I do see this page as being a catch-all for managing packages/dependencies. Once we need to differentiate between language specific and chip/OS specific managers, we can further flesh this out into separate pages. This page of documentation is under 50 lines so I think it's good enough for it to be as general as it is.

PR #298, opened 2022-10-27
Fix node version in deploy script

kleinschmidtj:
The deploy is failing. See [this error from the latest build](https://github.com/trussworks/Engineering-Playbook/actions/runs/3338593747):

```
[2/4] Fetching packages...
error @docusaurus/core@2.0.0-rc.1: The engine "node" is incompatible with this module. Expected version ">=16.14". Got "14.17.5"
error Found incompatible module.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
Error: Process completed with exit code 1.
```
This PR updates the node version in the deploy script to match the version in the `.node-version` file.

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -15,7 +15,7 @@ jobs:
 
       - uses: actions/setup-node@v3
         with:
-          node-version: 14.17.5
+          node-version: 16.16.0
 
       - name: Install yarn
         run: npm install yarn


.github/workflows/deploy.yaml
```diff
@@ -15,7 +15,7 @@ jobs:
 
       - uses: actions/setup-node@v3
         with:
-          node-version: 14.17.5
+          node-version: 16.16.0
```
rogeruiz, line 18: Love this solution. I wonder if it would make sense to read the version from the `.node-version` file rather than setting it here manually. @rpdelaney this is another good chance to reduce toil here.

kleinschmidtj: Yeah, that would make sense to me. I can look into it after we merge this PR.

rpdelaney: As far as I know, awareness of `.node-version` would have to be built-in to the `setup-node` action. From the [README](https://github.com/actions/setup-node#usage):
> The node-version input is optional. If not supplied, the node version from PATH will be used. However, it is recommended to always specify Node.js version and don't rely on the system one.
It sounds like we could remove this and it might do what we want, but that's discouraged. I don't know why this is discouraged.

PR #304, opened 2022-11-03
Add InfraSec ADR 0001 - Tests in Terraform

rpdelaney:
Add an ADR documenting a proposed decision to remove Terratests from our open source Terraform modules.


ralren:
I'm team remove tests.

jsclarridge:
Could you provide a link to a terratest that is currently or was recently broken? An example might help to illustrate the problem we're trying to solve.

clint-truss:
I'm in agreement that removed tests are far better than tests that give ambiguous signal. So, in that spirit, I agree with the ADR. But, as I read, I was wondering if there is space to maintain good tests for one or two high usage/high contribution modules. Just in the infra huddle a few days ago, three different projects listed the same three modules (S3, ALB, ECS). So, if we can determine which of those is our highest used/most contributed to module, then perhaps it would be worth it to develop good tests with alerting for when they break etc for that single module. Then over time, if we can get an OSS maintainer staffed, we have a model for what "good practice" looks like here. But until then, we don't have a huge maintenance burden.

rpdelaney:
> I was wondering if there is space to maintain good tests for one or two high usage/high contribution modules.

Without a plan for how we're going to maintain that, I worry that we'd just be downsizing this problem instead of eliminating it.

marycrawford (approved):
Nice that you created a template for future ADRs.  If we no longer need terratest then it is a good idea to remove it.  A separate conversation can happen regarding what test can or should be used to replace this at another time.  Perhaps an epic for the future.

clint-truss:
> > I was wondering if there is space to maintain good tests for one or two high usage/high contribution modules.
> 
> Without a plan for how we're going to maintain that, I worry that we'd just be downsizing this problem instead of eliminating it.

That's a fair point. Maybe just think of it as a "someday" wish. We do need to start with a plan for being able to maintain this stuff.

PR #303, opened 2022-11-02
Remove custom TOC from markdown

kleinschmidtj:
Removes the custom Table of Contents from markdown files. Docusaurus provides TOC out of the box.

Closes #254

PR #305, opened 2022-11-09
Add local hook to block mdformat TOCs

rpdelaney:

- Add local pre-commit hook to block mdformat TOCs. See discussion in PR #303


rpdelaney:
cc @rogeruiz @kleinschmidt

rpdelaney:
wtf why did I get stuck with these markdownlint fails >_<

PR #317, opened 2023-02-03
Adding a tip for folks reading through

rogeruiz:
With the updated InfraSec documentation in #308 about CI/CD, this larger
non-practice focused section needs a little tip in the beginning for any folks
reading. We can remove it once we've settled on things there.


rogeruiz:
Captura de pantalla 2023-02-03 a la(s) 6 29 14 p m

Here's what it looks like ☝🏽

PR #349, opened 2023-05-16
Rework CI/CD wishywashyness about prescriptions

rpdelaney:
We want to set standards. That means being prescriptive in a certain sense.

This change tries to clarify that.


esacteksab (commented):
:heart:

PR #398, opened 2026-06-12
ADR 0006 - Default to mise for tool dependency management

brandonlenz:
## https://trussworks.slack.com/lists/T0278L93W/F09HZLV0U4X?record_id=Rec0B9BJE1AAH

- Adds the ADR

I suppose it might also make sense to swap out .node-version in this branch to mise.toml + some docs updates, though I think I'd rather do that in its own PR


brandonlenz:
Ooof. We've go a misconfigured CI check (I think): https://github.com/trussworks/Engineering-Playbook/actions/runs/27444091585/job/81124889379?pr=398

edit: thanks @ronaktruss

docs/appeng/adrs/0006-default-to-mise-for-project-tool-version-management.md
```diff
@@ -0,0 +1,89 @@
...
+
+#### Pros
+
+- `+` Ubiquitous on most macOS machines, with an enormous package catalog.
+
+#### Cons
+
+- `-` Manages one version of each tool globally. Two engagements wanting different tool versions collide causing issues for practitioners splitting their time.
+- `-` A `Brewfile` declares version-agnostic packages. Reproducibility isn't the tool's goal.
+
```
felipe-lee, line 68: **suggestion**: also only on mac/linux. given we have projects with windows GFE, and, at least on some projects, we can't use WSL, that would mean this wouldn't work for those machines. We may also work with agencies/companies that default to having windows machines.

docs/appeng/adrs/0006-default-to-mise-for-project-tool-version-management.md
```diff
@@ -0,0 +1,89 @@
...
+While not selected for this specific decision, we're fond of Nix, so it stays on the table if reproducibility needs outgrow basic version pinning and ease of use.
+
+#### Pros
+
+- `+` Reproduces entire environments, not just tool versions.
+
+#### Cons
+
+- `--` _Steep_ learning curve.
```
felipe-lee: **suggestion**: I'd also add that at least on MilMove:

* tracking what versions of tools were available was a pain, to the point that Drew wrote a tool for us to use to find the right versions of things.
    * I haven't used it since, so maybe that's better now, but not sure.
* we would sometimes be bitten by an update to a tool we used taking a long time to land in the nix ecosystem.

docs/appeng/adrs/0006-default-to-mise-for-project-tool-version-management.md
```diff
@@ -0,0 +1,89 @@
+---
+title: Default to mise for tool dependency management
```
felipe-lee: **nit**: i think it might be better to call it `tool version management`

docs/appeng/adrs/0006-default-to-mise-for-project-tool-version-management.md
```diff
@@ -0,0 +1,89 @@
...
+It slots in to new and existing repos alike, and we can be as opinionated (or not) as we want about its use.
+Setup is fast, and the knowledge transfers between engagements instead of resetting with each one.
+
+## When to Not Implement This Decision
+
+- The client mandates their own dev tooling accepting no alternatives. We work with what they have.
+- The project needs hermetic, fully reproducible environments. That falls into Nix territory (See [nix README](../../developing/nix/README.md)).
+- We're not touching dev setup at all on the engagement. Don't introduce tooling that we won't help demonstrate the advantages of.
+
+## Alternatives Considered
```
felipe-lee: **suggestion**: I think it'd be good to have mise below as well, to have clear pros/cons in the same format as the other options.

docs/appeng/adrs/0006-default-to-mise-for-project-tool-version-management.md
```diff
@@ -0,0 +1,89 @@
...
+## Decision
+
+Default to [mise](https://mise.jdx.dev/) to manage tool dependencies on our projects.
+
+Mise covers every runtime we'll meet and respects whatever a client repo already uses: `.tool-versions`, `.nvmrc`, asdf plugins.
+Tools are pinned in one `mise.toml`, installed the same way locally and in CI, with no shims in the way.
+`which node` tells the truth and IDEs behave.
+
+Its task runner and env management are bonuses (and near drop-in replacements for Makefile and direnv), but they're not the basis for _this_ decision.
+
```
felipe-lee, line 25: **suggestion (blocking)**: this may go in the pros/cons section for mise if you want to add that section, but in general, we should probably note that mise:

* has one primary maintainer that relies a decent amount on AI afaik. it's not necessarily a deal-breaker, but is a risk we should include in our ADR.
* is a newer tool than the rest, so may not have been reviewed at Agencies we work in

docs/appeng/adrs/0006-default-to-mise-for-project-tool-version-management.md
felipe-lee: **thought**: I hadn't thought of this in the slack thread, but i wonder if we should have what is effectively a "do nothing" option.

felipe-lee (commented):
thanks for putting this together! i left some suggestions. most aren't blocking, but i do think we should add explicit risks/cons for mise

docs/appeng/adrs/0006-default-to-mise-for-project-tool-version-management.md
```diff
@@ -0,0 +1,89 @@
...
+As consultants, every engagement means a new codebase, a new stack, and a new pile of tools to install before we can run anything.
+We need one tool we can bring to any project that pins every runtime it needs, in one file, the same way on every machine.
+
+## Decision
+
+Default to [mise](https://mise.jdx.dev/) to manage tool dependencies on our projects.
+
+Mise covers every runtime we'll meet and respects whatever a client repo already uses: `.tool-versions`, `.nvmrc`, asdf plugins.
+Tools are pinned in one `mise.toml`, installed the same way locally and in CI, with no shims in the way.
+`which node` tells the truth and IDEs behave.
```
ronaktruss: Should we make this statement more generic and then provide this as an example? ie. Running `which  <tool name>`

docs/appeng/adrs/0006-default-to-mise-for-project-tool-version-management.md
```diff
@@ -0,0 +1,89 @@
...
+
+Because it doesn't care what the stack is. One tool works on a Node project, a Python project, a JVM project, or all three at once.
+It slots in to new and existing repos alike, and we can be as opinionated (or not) as we want about its use.
+Setup is fast, and the knowledge transfers between engagements instead of resetting with each one.
+
+## When to Not Implement This Decision
+
+- The client mandates their own dev tooling accepting no alternatives. We work with what they have.
+- The project needs hermetic, fully reproducible environments. That falls into Nix territory (See [nix README](../../developing/nix/README.md)).
+- We're not touching dev setup at all on the engagement. Don't introduce tooling that we won't help demonstrate the advantages of.
```
ronaktruss: Nit: reorder these two. `The project needs hermetic, fully reproducible environments.` is a rare requirement and imo least likely to be relevant to the reader

docs/appeng/adrs/0006-default-to-mise-for-project-tool-version-management.md
```diff
@@ -0,0 +1,89 @@
...
+### [asdf](https://asdf-vm.com/)
+
+#### Pros
+
+- `+` Mature, and its plugin ecosystem set the standard.
+- `+` The Go rewrite (0.16+) fixed the old performance complaints.
+
+#### Cons
+
+- `-` It's shim-based, and the indirection breaks `which` and can confuse IDEs, leading to potential developer swirl.
```
ronaktruss: The opposite of this point is also mentioned in the decision. It might be worth having an appendix section that double clicks into this explaining this issue in more detail. That way folks not familiar with the inner workings of these tools and these commands can at least understand the issue caused by this at a high level.

docs/appeng/adrs/0006-default-to-mise-for-project-tool-version-management.md
brandonlenz: Good call, the file is named that way already so I just crossed some wires apparently

docs/appeng/adrs/0006-default-to-mise-for-project-tool-version-management.md
brandonlenz: Do you have a good source I can include for the maintainer's reliance on AI? (I'll do a little research into this too, but I figured you probably have something handy)

docs/appeng/adrs/0006-default-to-mise-for-project-tool-version-management.md
brandonlenz: I tried to frame this (starting with the title) as "default" because this should not read "All projects must use Mise" and more "If you have a need for a tool version manager, use Mise if you can"

Rather than a do-nothing option, I'll soften the language at the start. That it'll read more like "If you need a tool version manager, then...", rather than right now which is more like "We work on lots of things we should always use a tool like this".

Kinda like how the postgres ADR doesn't have a do-nothing option: If you don't need a relational datastore, don't read/follow that ADR

docs/appeng/adrs/0006-default-to-mise-for-project-tool-version-management.md
brandonlenz: https://github.com/trussworks/Engineering-Playbook/pull/398/commits/8d130b7736043933ef97479e6801ced42cf1e107

docs/appeng/adrs/0006-default-to-mise-for-project-tool-version-management.md
brandonlenz: https://github.com/trussworks/Engineering-Playbook/pull/398/commits/78ced0261ce268e9fd0419bf8ea9d35e77bc9a49

docs/appeng/adrs/0006-default-to-mise-for-project-tool-version-management.md
brandonlenz: https://github.com/trussworks/Engineering-Playbook/pull/398/commits/bfa3e02f51663a74ffc6fbe2fcfc60217036d93a

docs/appeng/adrs/0006-default-to-mise-for-project-tool-version-management.md
brandonlenz: https://github.com/trussworks/Engineering-Playbook/pull/398/commits/b753df810ca0bdbb776d595da4ec994a06862d27

brandonlenz: "(and Linux)" because there are some Linux users who acknowledge that Homebrew works perfectly fine on Linux, but fervently maintain that there are much better options 🤪

docs/appeng/adrs/0006-default-to-mise-for-project-tool-version-management.md
brandonlenz: https://github.com/trussworks/Engineering-Playbook/pull/398/commits/b78d6764803da245e33986fc8576e2a6acfb9857

docs/appeng/adrs/0006-default-to-mise-for-project-tool-version-management.md
brandonlenz: Added in https://github.com/trussworks/Engineering-Playbook/pull/398/commits/d851230328588b71fc1792a82e98b45453f34988

docs/appeng/adrs/0006-default-to-mise-for-project-tool-version-management.md
brandonlenz: I added the other points more generically so far in https://github.com/trussworks/Engineering-Playbook/pull/398/commits/d851230328588b71fc1792a82e98b45453f34988

I would like to add a source or maybe specific framing about the AI portion. Our engineering playbook (including this repo) is a public artifact, and given the interest in AI from potential clients I don't want to give the impression that AI is a dealbreaker in all forms.

We can get behind well-reasoned and thoughtful/useful applications of AI.

docs/appeng/adrs/0006-default-to-mise-for-project-tool-version-management.md
brandonlenz: Added in https://github.com/trussworks/Engineering-Playbook/pull/398/commits/0b9244e3647a6e8a66804aa0b363dbb684002add

ronaktruss:
@brandonlenz et al,  Do you want to socialize this during our monthly engineering meeting later this week?

brandonlenz:
> @brandonlenz et al, Do you want to socialize this during our monthly engineering meeting later this week?

I'm happy to talk about it, sure!

brandonlenz:
For visibility, there's an alternative the folks might also prefer: https://github.com/trussworks/Engineering-Playbook/pull/407

edit merged in https://github.com/trussworks/Engineering-Playbook/pull/398/commits/0c20a33adf9d8611ee6643e05832ecfffc081530

docs/appeng/adrs/0006-default-to-mise-for-project-tool-version-management.md
felipe-lee: I'd mostly noted that the repo seems to have AI workflows. I did a search to see if he's written about it, and it seems like this transcript mentions it: https://www.devtools.fm/episode/173?view=transcript

I agree with your point though about AI not being a dealbreaker, and it seems like he's being careful with it from the skim of that transcript i did, rather than just relying on it without any checking. I'm good with leaving that out and just mentioning the single maintainer risk.

esacteksab:
Not sure where to put this, but wanted to bring attention to `mise` and `1password` friction https://github.com/jdx/mise/discussions/3542 -- the integration between the two isn't great. `direnv` IME is better. Also `make` is native on most of our systems, `mises` task runner is another DSL. I'll add more, but wanted to drop this here. I was a big proponent of `asdf` vs `nix` but abandoned `asdf` with their Go rewrite and how they fumbled the rollout. While I use `mise` -- it isn't as well refined as `asdf` in my experience. There are some odd edge cases with `.tool-version` vs `.mise.toml` or `.mise.local.toml` (which I ignore in my global `.gitignore`). I often find myself oscilating back and forth between `.tool-version` and `.mise.toml.` for version management. FWIW, Dependabot doesn't support `.mise.toml` but does support `.tool-version` so if you want to manage those updates, you can't use native `mise` config file. `direnv` and `.env` handles injecting environment variables and enabling virtualenvs a lot better than `mise` IME.

More to follow when I have some more time.

ieffendi-twks:
@esacteksab has some great points that aligned with my early experience with `mise`. I’d say that over the last ~6 months, the tool has matured and my workflow has shifted, addressing some of the initial friction that I’ve run into. I’m using this space to discuss some of those specific patterns as they might be useful for new adopters (and can help inform a decision on this ADR).
### Re: `1password` friction
The problem raised ~2 years ago https://github.com/jdx/mise/discussions/3542 on the discussion thread Barry shared is a real limitation. Since then, the creator’s answer has been [fnox](https://fnox.jdx.dev/):

> [mise](https://mise.jdx.dev/) has support for [encrypted secrets](https://mise.jdx.dev/environments/secrets/) but mise's design makes it a poor fit for remote secrets. mise reloads its environment too frequently—whenever a directory is changed, mise x is run, a shim is called, etc. Any other use-case like this mise leverages caching but secrets are an area where caching is a bad idea for obvious reasons. It might be possible to change mise's design to retain its environment in part to better support something like this but that's a huge challenge.
> 
> Basically it's just too hard to get remote secrets to work effectively with mise so I made this a standalone tool.
> -- [Why is this a standalone CLI and not part of mise?](https://fnox.jdx.dev/guide/what-is-fnox.html#why-is-this-a-standalone-cli-and-not-part-of-mise)

**I agree that `mise` is not the right tool for encrypted secrets** – but I am also hesitant to fall into the trap of treating `mise` like it can solve everything. When you only have a hammer… Instead, for the last few months on personal projects, I’ve been using `fnox`, and its first-class support for 1Password makes it worth considering as a companion tool. Unlike `mise`, which reloads its environment aggressively on every directory change and command invocation, `fnox` is designed to be called explicitly—when you need secrets, you ask for them. Because `fnox` is intentionally separate from `mise`, it also avoids the caching and reload problems that make secret management inside `mise` unreliable. You do not need to `fnox` to use `mise` (or vice versa), but that discussion might be out-of-scope for this ADR.
### Re: `.tool-versions` vs. `mise.toml`
From the maintainer’s standpoint, `.tool-versions` is not the recommended approach when using `mise`. If we adopt `mise` on projects going forward, we should accept the trade-offs mentioned, including the lack of Dependabot support. It is worth calling out that [efforts to add](https://github.com/dependabot/dependabot-core/pull/14500) official `mise` support to Dependabot have stalled as `jdx` cannot commit to long-term ownership over the integration with Dependabot https://github.com/dependabot/dependabot-core/pull/14500#issuecomment-4172839552.

In an effort to seriously consider what workarounds might exist, I popped a question into a quick AI chat and got back a couple options:
- One is simply maintaining a `.tool-versions` as a canonical version declaration and using `mise` for everything else (tasks, environments, settings configuration). Dependabot can natively update `.tool-versions`, but it adds another file to maintain and we lose `mise.toml`-specific features (like `path://` and `source://` backends) for those tools. Nothing stops us from mixing and matching as needed.
- Another is adding a `CI` job that can extract the `[tools]` table from `mise.toml` to write out a temporary `.tool-versions` file. YAML and JSON CLI parsers like `yq` and `jq` are already compatible with `*.toml` files so this is a one-line command: `yq -r '.tools | to_entries | .[] | "\(.key) \(.value)"' mise.toml > .tool-versions`. Dependabot gets what it needs and `mise.toml` acts as a single source of truth.
- Pinning versions directly in the CI would allow Dependabot to update them as it has access to our pipeline definitions (`ci.yml`). This one risks drift between our local versions (`mise`) and what’s in our `CI`. In fairness, not a fan of this approach, but including it since it was recommended.
- Worth mentioning `Renovate` allows for custom tool manager support - not a realistic solution since we rely on Dependabot as our tool of trade but perhaps worth keeping in mind if that ever gets discussed.

### Re: DSL complexity for `mise` tasks
There is a ton of different ways to write `mise` tasks and an equivalent risk of sprawl if we do not standardize on an approach. I have gravitated towards using executable shell scripts under a `.mise-tasks/` directory co-located with a project’s `mise.toml` configuration. This has let me avoid defining complex tasks inside of the `mise.toml` itself.

Your typical shell script would just get a short, optional preamble to configure the task for `mise`. The `mise`-specific header is comment-based, so a well-defined shell script can just be executed as normal.

```bash
#!/usr/bin/env bash
#MISE description="This is a typical shell script"
#USAGE arg "<example>" help="An example positional argument."

set -euo pipefail # Can still set shell options like a normal file.

echo "Arguments are made available with the 'usage_' prefix: usage_example ==> ${usage_example}"
```

The other hidden benefit is that the shebang operator can be used for other languages such as `node` and `python`:

```js
# example.js
#!/usr/bin/env node
//MISE description="Hello, World in Node.js"

console.log("Hello, World!");
```

```python
# example.py
#!/usr/bin/env -S uv run --script
#MISE description="Hello, World in Python."

print('Hello, World!')
```

If a task is a one-liner (e.g., `npm run test ...`) with some complex set of arguments that you don't want to type out repeatedly, [TOML task syntax](https://mise.jdx.dev/tasks/toml-tasks.html) is a good fit. For anything that requires multiple lines, the task file approach is very similar to setting up a `./bin/<task>` file and making it executable. As someone who maintains a lot of shell-based automation in my spare time, I find this pattern keeps it readable.

There is also support for a [usage-spec DSL](https://mise.jdx.dev/tasks/task-arguments.html#complete-usage-specification-reference) for declaring arguments, options, and defaults. I find this syntax *incredibly* efficient condensing complex, [difficult-to-read `getopts` preambles](https://www.redhat.com/en/blog/arguments-options-bash-scripts#handling-options) in shell scripts into simple [`#USAGE arg`](https://mise.jdx.dev/tasks/task-arguments.html#file-task-headers) pragma headers.

In my view, a task manager should get out of your way. A few thoughts on the common alternatives:
- `make` and its `.PHONY:` targets are ubiquitous but the inherently sparse syntax does not lend itself to easy comprehension, especially for those less familiar with the format. `make` has a specific purpose as a *build system* and that while it can be coerced into executing tasks, it wasn’t designed with that purpose. 
- I've used [`just`](https://github.com/casey/just), a command runner that cribs its syntax from `make`, and I have the same gripe: familiarity with `make` syntax is driving the interface design instead of designing an interface for the specific purpose of task management.
- [`task`](https://taskfile.dev/) is another cross-platform command runner alternative that provides useful features behind a YAML-file-based implementation. CI/CD engineers will feel right at home with the YAML format and there are some neat features here. That said, I often reach for `mise`'s file-based scripts when complexity arises as I find shell scripts to be more expressive than YAML files.

### Re: `uv` and `virtualenv` support
I have found `mise`'s support for virtual environments for both `uv` and non-`uv` Python projects to be pretty seamless. It has definitely matured over the past few months. The docs have a section dedicated to [automatic virtualenv activation](https://mise.jdx.dev/lang/python.html#automatic-virtualenv-activation) and the mechanisms for using it are well-documented. That said, I'm not 100% familiar with Truss' general approach for Python usage. I personally gravitate towards `uv` and `pyproject.toml` configurations over the older `pip` / `wheel` / `setuptools` approach. Maybe this deserves a *separate* ADR but I prefer `uv`'s unified and performant approach over [`distutils` and `setup.py`](https://the-hitchhikers-guide-to-packaging.readthedocs.io/en/latest/index.html), [`pip` + `venv`](https://packaging.python.org/en/latest/guides/installing-using-pip-and-virtual-environments/), [`PDM`](https://pdm-project.org/en/latest/), [`hatch`](https://hatch.pypa.io/latest/), and [`poetry`](https://python-poetry.org/). [The packaging ecosystem in 2026 is nauseatingly fractured for Python](https://repoforge.io/blog/posts/the-state-of-python-packaging-in-2026/).

docs/appeng/adrs/0006-default-to-mise-for-project-tool-version-management.md
```diff
@@ -0,0 +1,86 @@
+# 0006 – Default to mise for project tool version management
```
ieffendi-twks, line 1: If this ADR is for specifically "tool version management", I think it is worth including companion ADRs to cover task management and environment management. Much of the value `mise` brings is through the consolidation of several capabilities into one tool (and single DSL) structure.

There are some pretty big impacts `mise` has on environment variable resolution that are inseparable from this choice to use it as a tool version manager. Projects using `direnv`, for example, would not be compatible with mise out of the box. That alone is worth a callout in this ADR, if not in a separate one. Especially since `mise` relies on shell hooks and shims.

@esacteksab made a great point about the 1Password friction; so I'd also take it a step further and say a section needs to highlight that weakness (and any workarounds / mitigations) since 1Password is a defacto stable of Truss' working tools. If I have time, I'd throw together an ADR for `fnox` to complement that weakness.

docs/appeng/adrs/0006-default-to-mise-for-project-tool-version-management.md
```diff
@@ -0,0 +1,86 @@
...
+- `-` Newer than the alternatives and less battle-tested, though Trussels have used it for several years.
+- `-` One primary maintainer, a potential bottleneck despite strong community engagement.
+
+### [asdf](https://asdf-vm.com/)
+
+- `+` Language and framework agnostic.
+- `+` Mature, with the plugin ecosystem that set the standard.
+- `-` [Shim-based](#what-are-shims); the indirection breaks `which` and can confuse IDEs.
+
+### Language-specific managers ([nvm](https://github.com/nvm-sh/nvm), [pyenv](https://github.com/pyenv/pyenv), [sdkman](https://sdkman.io/), et al.)
```
ieffendi-twks, line 54: In my practice, I've found that `mise` as a polyglot version manager is *especially* useful for managing development dependencies and infrastructure tooling that is not *directly* related to the project at hand and it is positively complimented with a specific-language manager.

It's actually one of the supported pathways with using [`mise` + `uv` for Python scripts and package development](https://mise.jdx.dev/mise-cookbook/python.html#mise-uv). 

To that end, I am not sure it serves as a total replacement for all of these tools, only for other polyglot version managers (like SDKMAN). Where `mise` shines is providing a consistent, stable configuration that can be shared across machines with hierarchical directory management. In a large monorepo, a `mise.toml` in the root can go a long way without getting in the way of any subdirectories that have specific requirements. They can have their own nested `mise.toml`'s that layer environments well and have language-specific version managers for the area of overlap that isn't consistent between projects.

docs/appeng/adrs/0006-default-to-mise-for-project-tool-version-management.md
```diff
@@ -0,0 +1,86 @@
...
+
+## When Not to Apply This Default
+
+This is a default, not a mandate. It does not apply when:
+
+- A client mandates their own tooling with no alternatives. We work with what they have.
+- The engagement does not touch dev setup. Do not introduce tooling we will not help demonstrate.
+- The project needs hermetic, fully reproducible environments. That is [Nix](#nix) territory (see [nix README](../../developing/nix/README.md)).
+
+## Options Considered
```
ieffendi-twks, line 37: Tool version management, to me, is about declaratively managing a working execution environment. 

With that context, I would actually see these as alternatives to consider as well:
- [Devcontainers](https://containers.dev/) (heavyweight)
- [Docker Compose](https://docs.docker.com/compose/) (although there might be some licensing concerns with Docker)
- [`devenv`](https://devenv.sh/) (Nix-based reproducible envs; I find its DSL and nix's DSL to be less elegant / less efficient)
- [`proto`](https://moonrepo.dev/proto) (paid polyglot version manager); worth also mentioning [`moon`](https://moonrepo.dev/moon) by the same devs.
- [`pixi`](https://pixi.prefix.dev/latest/python/tutorial/) (but honestly might be superseded by `uv`)
- [`Bazel`](https://bazel.build/) (complex, fast, hermetic build system - but it's not really applicable to our projects).
- [`Earthly`](https://earthly.dev/) (so Dockerfile + Makefile put into a blender. honestly a worthy contender as a build tool + version manager).

brandonlenz:
@esacteksab @ieffendi-twks super helpful details regarding how you all are working with these mise.

I do want to keep this adr constrained to tool version management, so I'll avoid introducing pros and cons about environment, secrets, and task management to this ADR, even if they are very real points of friction. Since these are things that alternatives like `asdf` also don't do, comparing [asdf + other tools] to [mise + other tools OR just mise] isn't consequential here.

The dependabot limitation is definitely a con that should be listed here, where asdf gets an automation boost via dependabot for keeping tools up-to-date, where mise does not (without workarounds). Admittedly, I've only ever managed tool version manually, since the projects I've used it on sat happily with a small-ish number of LTS tools.

I'll add that to the ADR!

docs/appeng/adrs/0006-default-to-mise-for-project-tool-version-management.md
ieffendi-twks, line 1: I think we're missing a big section on CI/CD integration. If `mise.toml` is used to manage our tool versions in our local environment, how should we approach communicating those versions to our CI/CD pipeline? It's a similar question posed by the lack of native Dependabot support for `mise.toml`. If we don't have a good answer, we risk version drift.

In my mind, our options are:
- Install `mise` as part of our CI pipeline on projects where `mise` is used and use `mise install` to hydrate the necessary tools. Optionally: a `mise.ci.toml` can be used for specifically installing a subset or superset of the necessary tooling. This could be a custom step or a GitHub Action.
- Maintain a parallel version declaration for CI-native setup actions.
- Write a custom script that reads the `mise.toml` and outputs a version file our CI pipeline can utilize.

docs/appeng/adrs/0006-default-to-mise-for-project-tool-version-management.md
```diff
@@ -0,0 +1,86 @@
...
+### [Nix](https://nixos.org/)
+
+Not selected here, but kept on the table if reproducibility needs outgrow version pinning.
+
+- `+` Reproduces whole environments, not just tool versions.
+- `-` Tool availability within the Nix ecosystem has been hard to determine in past use.
+- `-` New tools and versions land on an unpredictable cadence, which complicates version management.
+- `--` Steep learning curve.
+
+## Appendix
```
ieffendi-twks, line 78: Would it be worth adding a section here about `mise`'s IDE support? There are some maintained plugins for `mise` and `mise tasks` that can be included.

docs/appeng/adrs/0006-default-to-mise-for-project-tool-version-management.md
ieffendi-twks: I think we should add a section to the appendix that addresses the single-maintainer risk. Notably with information that helps suggest its long-term stability (e.g., approximately ~600 releases since the first in [August 2023](https://github.com/jdx/mise/releases/tag/v2023.8.0)) and possible consequences of supply-chain risk (e.g., if a project needs to un-adopt mise due to it falling out of favor, however unlikely, what would the damage be?).

docs/appeng/adrs/0006-default-to-mise-for-project-tool-version-management.md
```diff
@@ -0,0 +1,86 @@
...
+- `-` N stacks means N tools per machine, reinstalled per engagement, with no single manifest to declare them.
+
+### [Homebrew](https://brew.sh/)
+
+Homebrew installs tools; it does not pin them per project. We keep using it for that (including to install mise itself).
+
+- `+` Ubiquitous on macOS, with a large package catalog.
+- `-` Manages one global version per tool. Two engagements needing different versions collide.
+- `-` A `Brewfile` pins packages, not versions; reproducibility is not its goal.
+- `-` macOS and Linux only, so it does not cover projects that require Windows (often for GFEs).
```
ieffendi-twks, line 67: Is it worth explicitly calling out under `mise` that `mise` supports Windows natively (via Git Bash, MSYS2, and PowerShell) as well as on WSL2? The GFE point seems relevant to our working scenarios.

docs/appeng/adrs/0006-default-to-mise-for-project-tool-version-management.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Options Considered
+
+### [mise](https://mise.jdx.dev/)
+
+- `+` Language and framework agnostic.
+- `+` Reads existing `.tool-versions`, `.nvmrc`, and asdf plugins, easing migration and per-project adoption.
+- `+` Rewrites `PATH` to point at real binaries, avoiding the [shim](#what-are-shims) approach.
+- `+` Bundles documented extras beyond tool management ([tasks](https://mise.jdx.dev/tasks/), [environments](https://mise.jdx.dev/environments/)).
+- `-` Newer than the alternatives and less battle-tested, though Trussels have used it for several years.
```
ieffendi-twks, line 45: Additional concern around the trust/security model for `mise` tools. `mise` has the option to compile some dependencies from source (if you configure it that way) but it defaults to downloading pre-compiled binaries. Mitigating the risk of supply-chain compromised downloads will take additional steps on our part, either through the practice of having minimum release ages / pinned versions and an external scanner like [`trivy`](https://trivy.dev/) / [`grype`](https://github.com/anchore/grype) / [`bandit`](https://github.com/PyCQA/bandit) / etc., or by having better SLSA verification beyond checksums.

Some backends (like `aqua` for github tools) have native software verification in place (and it can actually be configured) but others do not have the same: https://mise.jdx.dev/security.html

There's support for top-level version filtering and transitive dependency filtering during installation but it's backend-dependent (e.g., you can get some in `npm:` but not the same filtering support in `cargo:`).

Worth calling this out either in the appendix or as a benefit under the `mise` tool choice.

ieffendi-twks (changes requested):

docs/appeng/adrs/0006-default-to-mise-for-project-tool-version-management.md
esacteksab: Will GFE's have Git Bash, MSYS2 and Powersell or WSL enabled?

brandonlenz: Definitely not always. On my most recent project, Git Bash was something you needed to go through an administrator install process for. I just wanted OpenSSL 😢

PR #399, opened 2026-06-15
Update vale settings

ronaktruss:
## Update local and ci vale settings

- Updates ci to point to newly located repo ([ref](https://github.com/vale-cli/vale-action))
  - And updates the version
- Updates local vale.ini settings to account for changes in v3 ([ref](https://vale.sh/docs/keys/vocab#folder-structure))


ronaktruss:
@brandonlenz I think this should resolve the issue you are seeing in https://github.com/trussworks/Engineering-Playbook/pull/398

brandonlenz:
image

Huh, does it work with just `@v2`? Or maybe best to pin it to the 2.1.2 sha: `85f9f7f2c5f449ac0ae5b66662961bae3f77ca6a`

ronaktruss:
> image
> 
> Huh, does it work with just `@v2`? Or maybe best to pin it to the 2.1.2 sha: `85f9f7f2c5f449ac0ae5b66662961bae3f77ca6a`

Yeah tagging with @v2 is likely the best bet, but I'm not sure what's causing the error now([ref](https://github.com/trussworks/Engineering-Playbook/actions/runs/27574996454/job/81520949049?pr=399)). The original issue with the upgrade was they didn't include the `v` in the updated tag.

I'm going to just update this to @v2 which still points to the same sha as v2.1.1, but should be updated if they decide to align their tagging nomenclature.

ronaktruss:
If at some point v2 does point to 2.1.2 without any additional changes in the corresponding repo, we'll need to dig into the error.

brandonlenz (approved):
Thank you! LGTM 🫡 

If it breaks again, should be no biggie to fix, but hopefully that would come with a corresponding major semver update 🤷🏻

PR #404, opened 2026-06-22
Update settings

ronaktruss:
## Fix failing vale actions

- Updates settings to ensure the reporter is consistent with when we are running the action (it defaults to github-pr-annotations and we are only running this on PRs)

### DIFF
--- a/.github/workflows/check_prose.yaml
+++ b/.github/workflows/check_prose.yaml
@@ -11,5 +11,7 @@ jobs:
 
       - name: Vale linter
         uses: vale-cli/vale-action@v2
+        with:
+          reporter: github-check
         env:
           GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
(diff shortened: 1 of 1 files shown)


ronaktruss:
This should resolve the failing test in https://github.com/trussworks/Engineering-Playbook/pull/404

.github/workflows/check_prose.yaml
```diff
@@ -11,5 +11,7 @@ jobs:
 
       - name: Vale linter
         uses: vale-cli/vale-action@v2
+        with:
+          reporter: github-check
```
brandonlenz: I wonder if we should run this only on PRs and then use gihub-pr-check.

Not blocking, as I can merge main on #398 to test how this works.

brandonlenz (approved):

.github/workflows/check_prose.yaml
ronaktruss: Yeah that's a better solution

PR #407, opened 2026-06-25
Optional Mise ADR rewrite

brandonlenz:
Here's a ~rewrite of the ADR using a template and style choices that _I_ personally prefer.

Figured this would be easier than clobbering the existing PR in case.

If we prefer this, we can just merge into the base branch. If we'd rather preserve consistency with _some_ of the other ADRs in this repo, we can close this.


felipe-lee (approved):
i personally prefer this format too, and don't think sticking to the existing format of past ADRs should necessarily stop us from changing it going forward.

